PR #9, opened 2018-08-01
Added some javadoc to S7 communication path and several todos that ma…

JulianFeinauer:
Hey all,

i went through the S7 Communication path to understand it for myself and by doing that I tried adding useful javadoc at several classes.
Furthermore, I added several TODOs where I observed possible problems.

I am not sure if all of this javadoc is right so before merging this PR someone should go through it carefully and check if the things I state are right.

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadRequestItem.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadRequestItem.java
@@ -22,6 +22,18 @@ import org.apache.plc4x.java.api.model.Address;
 
 import java.util.Objects;
 
+/**
+ * Encapsulats one {@link RequestItem} that could be read multiple times, i.e., from the given Address on
+ * {@link ReadRequestItem#size} number of Items with equal datatype are read.
+ *
+ * Thus,
+ * <pre>
+ *     new ReadRequestItem(Int.class, adress, 5)
+ * </pre>
+ * basically reads 5 consecutive integers starting at the given {@link Address}.
+ *
+ * @param <T> Generic Type of expected Datatype.
+ */
 public class ReadRequestItem<T> extends RequestItem<T> {
 
     private final int size;
@@ -52,6 +64,8 @@ public class ReadRequestItem<T> extends RequestItem<T> {
             return false;
         }
         ReadRequestItem<?> that = (ReadRequestItem<?>) o;
+        // TODO 01.18.18 jf: we should also call the comparison of super at least otherwise this can lead to unwanted behavior.
+        // Perhaps we should generate a UUID or something for each ReadRequest to have a good equality comparison
         return size == that.size;
     }
 
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadResponseItem.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadResponseItem.java
@@ -24,6 +24,12 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
+/**
+ * Response to a {@link ReadRequestItem}.
+ * Can contain a list of values if the size in {@link ReadRequestItem} is larger zero.
+ *
+ * @param <T>
+ */
 public class ReadResponseItem<T> extends ResponseItem<ReadRequestItem<T>> {
 
     private final List<T> values;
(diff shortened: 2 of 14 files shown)


plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadRequestItem.java
```diff
@@ -52,6 +64,8 @@ public boolean equals(Object o) {
             return false;
         }
         ReadRequestItem<?> that = (ReadRequestItem<?>) o;
+        // TODO 01.18.18 jf: we should also call the comparison of super at least otherwise this can lead to unwanted behavior.
```
chrisdutz: Aren't we doing exactly this super check 4 lines above?

plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadRequestItem.java
```diff
@@ -52,6 +64,8 @@ public boolean equals(Object o) {
...
+        // Perhaps we should generate a UUID or something for each ReadRequest to have a good equality comparison
```
chrisdutz: Generation of UUIDs takes a lot of CPU and consumes Memory

plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadRequestItem.java
JulianFeinauer: I'm sorry, i really completely missed that... you're right.

plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/items/ReadResponseItem.java
```diff
@@ -60,6 +66,7 @@ public boolean equals(Object o) {
             return false;
         }
         ReadResponseItem<?> that = (ReadResponseItem<?>) o;
+        // TODO 01.08.18 jf: should we also use a UUID here or at least a check for equal request as this could lead to effects due to unwanted equality (same for hashCode).
```
chrisdutz: Generation of UUIDs takes a lot of CPU and consumes Memory

plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/netty/Plc4XS7Protocol.java
```diff
@@ -54,6 +54,17 @@ Licensed to the Apache Software Foundation (ASF) under one
 import static org.apache.plc4x.java.s7.netty.util.S7TypeDecoder.decodeData;
 import static org.apache.plc4x.java.s7.netty.util.S7TypeEncoder.encodeData;
 
+/**
+ * This layer transforms between {@link PlcRequestContainer}s {@link S7Message}s.
+ * And stores all "in-flight" requests in an internal structure ({@link Plc4XS7Protocol#requests}).
+ *
+ * TODO 01.08.18, jf: is a Cleanup of the requests map necessary? Can it occur that for a request no response is generated that leads to success or failure?
```
chrisdutz: Yes we should have some housekeeping mechanism in place that handles timeouts and connection problems. The driver implementation we are using in the Enthernet/IP driver (https://github.com/digitalpetri/ethernet-ip) has been a great inspiration on how to do this. We just haven't done it yet. Perhaps it would be a good idea to create a Jira for this?

plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/netty/model/params/VarParameter.java
```diff
@@ -24,6 +24,13 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 import java.util.List;
 
+/**
+ * "Command" Message to inform PLC of wanted operation.
+ * Also contains {@link VarParameter#items} that hold detailed information on the "Command", e.g.,
+ * addresses to read or to write to.
+ *
+ * TODO 01.08.18 jf: Could this be renamed to Something like Command as this seems to be the command message pattern?
```
chrisdutz: These classes are driver-internal classes. No user should ever deal with them. The naming is correlated to names used in some of the documentation of the S7 protocol. So even if I agree that the correct name for this class from a pattern point of view might be different, I would suggest to keep it the way it is, because this way there is the correlation to the informal specs.

JulianFeinauer: Okay, the correlation makes sense.
But its good to know that my assumption was right.
It simply helped my understanding of the process to see them as that.

PR #10, opened 2018-08-02
Fixed NPE in S7PlcConnection#close.

JulianFeinauer:
Fix for the NPE in S7PlcConnection's close method.
After fixing the NPE there was a netty blockingException, thus I had to refactor the program logic a bit.
Please check if this is suitable the way I dit this (I tried to add enough comment to make it understandable).

### DIFF
--- a/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/connection/S7PlcConnection.java
+++ b/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/connection/S7PlcConnection.java
@@ -19,6 +19,8 @@ under the License.
 package org.apache.plc4x.java.s7.connection;
 
 import io.netty.channel.*;
+import org.apache.commons.configuration2.Configuration;
+import org.apache.commons.configuration2.SystemConfiguration;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.plc4x.java.api.connection.PlcReader;
 import org.apache.plc4x.java.api.connection.PlcWriter;
@@ -56,6 +58,10 @@ import java.util.regex.Pattern;
 public class S7PlcConnection extends AbstractPlcConnection implements PlcReader, PlcWriter {
 
     private static final int ISO_ON_TCP_PORT = 102;
+
+    // Fetch values from configuration
+    private static final Configuration CONF = new SystemConfiguration();
+    private static final long CLOSE_DEVICE_TIMEOUT_MS = CONF.getLong("plc4x.s7connection.close.device,timeout", 1_000);
 
     private static final Pattern S7_DATABLOCK_ADDRESS_PATTERN =
         Pattern.compile("^DATA_BLOCKS/(?<blockNumber>\\d{1,4})/(?<byteOffset>\\d{1,4})");
@@ -175,41 +181,6 @@ public class S7PlcConnection extends AbstractPlcConnection implements PlcReader,
         return paramMaxAmqCallee;
     }
 
-    /**
-     * Closes the S7 connection.
-     * First, a disconnect request is send to the PLC and after that, the Session and the channel are closed and
-     * invalidated.
-     *
-     * Then {@link ChannelFuture#await()} is used to see if the PLC closes the connection as expected.
-     * We have to use {@link ChannelFuture#await()} here for two reasons.
-     * First, netty does not allow blocking calls inside {@link io.netty.util.concurrent.GenericFutureListener}.
-     * Second, a timeout ensures that the close operation will abort in case of, e.g., transportation problems and the
-     * jvm can shut down gracefully.
-     */
-    @Override
-    public void close() {
-        if ((channel != null) && channel.isOpen()) {
-            // Send the PLC a message that the connection is being closed.
-            DisconnectRequestTpdu disconnectRequest = new DisconnectRequestTpdu(
-                (short) 0x0000, (short) 0x000F, DisconnectReason.NORMAL, Collections.emptyList(),
-                null);
-            ChannelFuture sendDisconnectRequestFuture = channel.writeAndFlush(disconnectRequest);
-            // Wait if the PLC closes the connection
-            try {
-                // TODO 02.08.18 jf: Do we have global constants for things like timeouts?
-                channel.closeFuture().await(1_000);
-            } catch (InterruptedException e) {
-                logger.warn("Connection was not closed by PLC, has to be closed from driver side now.", e);
-            } finally {
-                // close the session itself.
-                channel.eventLoop().parent().shutdownGracefully();
-                super.close();
-            }
-        }
-
-    }
-
-
     @Override
     public Address parseAddress(String addressString) throws PlcException {
         Matcher datablockAddressMatcher = S7_DATABLOCK_ADDRESS_PATTERN.matcher(addressString);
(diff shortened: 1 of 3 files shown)


plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/connection/S7PlcConnection.java
```diff
@@ -183,14 +194,19 @@ public void close() {
                 (short) 0x0000, (short) 0x000F, DisconnectReason.NORMAL, Collections.emptyList(),
                 null);
             ChannelFuture sendDisconnectRequestFuture = channel.writeAndFlush(disconnectRequest);
-            sendDisconnectRequestFuture.addListener((ChannelFutureListener) future -> {
-                // Close the session itself.
-                channel.closeFuture().await();
+            // Wait if the PLC closes the connection
+            try {
+                // TODO 02.08.18 jf: Do we have global constants for things like timeouts?
```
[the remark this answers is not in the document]
JulianFeinauer: fixed and removed todo.

plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/connection/S7PlcConnection.java
```diff
@@ -183,14 +194,19 @@ public void close() {
...
+                channel.closeFuture().await(1_000);
+            } catch (InterruptedException e) {
+                logger.warn("Connection was not closed by PLC, has to be closed from driver side now.", e);
```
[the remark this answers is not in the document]
JulianFeinauer: done!

plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/connection/S7PlcConnection.java
```diff
@@ -183,14 +194,19 @@ public void close() {
...
+            } finally {
+                // close the session itself.
                 channel.eventLoop().parent().shutdownGracefully();
```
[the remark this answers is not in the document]
JulianFeinauer: Finally was unnecessary here so removed.

JulianFeinauer:
I incorporated fixes for all comments and tried to fix the failing IT (how can i run them locally?).
As the close in the IT throws a NPE I decided not to change the close method but to add try-catch to the IT as this is due to the ITs setup.

plc4j/protocols/s7/src/test/java/org/apache/plc4x/java/s7/connection/S7PlcConnectionIT.java
```diff
@@ -58,7 +58,12 @@ public void setUp() {
     @After
     public void tearDown() {
         if(s7PlcConnection.isConnected()) {
-            s7PlcConnection.close();
+            try {
+                s7PlcConnection.close();
+            } catch (NullPointerException e) {
```
chrisdutz: We should investigate this in the future and try to find out why the NPE is thrown and fix things, but I don't have any objections to this for now.

JulianFeinauer: My bet is that its this line here: channel.eventLoop().parent().shutdownGracefully();
Perhaps the embedded channels Embedded Event Loop has no parent?

sruehl: maybe we should wrap all shutdown/close calls with something like that: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#closeQuietly(java.io.Closeable...).
Do you still have the original stacktrace?

JulianFeinauer: The Stacktrace came from travis CI here https://travis-ci.org/apache/incubator-plc4x/builds/411177329

sruehl: route cause of failing test is `io.netty.util.concurrent.BlockingOperationException` and only NPE gets caught so the test still fails

JulianFeinauer: I added a fix for that.
Before catching the NPE an NPE was raised (trace from previous build:
java.lang.NullPointerException
	at org.apache.plc4x.java.s7.connection.S7PlcConnectionIT.tearDown(S7PlcConnectionIT.java:61))
Now as the NPE is catched a BlockingOperationException is thrown... weird.

sruehl: this is thrown on master as well, I justed wrapped the code like this:
```
 sendDisconnectRequestFuture.addListener((ChannelFutureListener) future -> {
                // Close the session itself.
                try {
                    channel.closeFuture().await();
                    channel.eventLoop().parent().shutdownGracefully();
                } catch (Exception e) {
                    e.printStackTrace();
                }

            });
```

JulianFeinauer: Yes, this is the code i refactored. Reason for the BlockingException is, that Netty does not allow the "await" call in a lambda which is excuted on its thread pool.
Thus, I changed program flow to avoid this.

sruehl: seems like the refactoring didn't solve this as this is still a netty thread? At least @chrisdutz is getting the very same exception on your branch right now... We should investigate why this

channel.closeFuture().await();
channel.eventLoop().parent().shutdownGracefully();
was there in the first place.

JulianFeinauer: @chrisdutz how is your setup?
For me, this removed the exception here for me with my S7.

chrisdutz: Hi Julian ... I just updated master ... could you check if this solves your problem?

JulianFeinauer:
@chrisdutz your fixes in the master made it work for me also.
As I found a solution to avoid the blocking exception in the meanwhile I combined your code with mine.
I think the only obvious solution is, as i propose, to use a CountDownLatch to communicate between the ChannelCloseFuture and the close method on another ThreadPool than the EventLoopGroup (which netty forbids).
Is this solution feasible for you?

sruehl:
one note at this point. As far as I understand netty throws this exception to indicate that it dislikes when the thread gets blocked. By using @chrisdutz or @JulianFeinauer code we might just trick the detection of netty. Im not quite sure but I think this is then more a hack/workaround rather then the intended use of netty. What I found what should come close to the intended behavior ist [this](https://netty.io/4.0/api/io/netty/util/concurrent/SingleThreadEventExecutor.html#addShutdownHook-java.lang.Runnable-). Maybe we should ask on the netty list how it is expected to do such things...

JulianFeinauer:
@sruehl my understanding is that netty avoids a scenario where one uses the event loop executors threads for blocking and thus, risks a deadlock. This should not be the case in this situation because the EventLoops Thread wich does the listener notification only counts down the latch.

But as this solution seems uglier than it should be it is possibly a good idea to go over to the netty list and ask there (I am totally new to netty).

chrisdutz:
Well I Ave started working on an option that works exactly the same way we are doing the connection, by creating a CompletableFuture and commuting this in a channel.closeListener ... Will finish that option tomorrow and have a look at your proposal tomorrow as well. Think we are all learning more and more with stuff like this ;-)

chrisdutz:
Ok ... so I just checked in the version I was talking about. It is basically structurally equal to the way we are doing things in the connection part. Also I just had a look at your change ... seems we are doing the same thing, but I would like to stick to CompletableFutures as your option is just doing the same but with a different syntax.

chrisdutz:
And I don't agree with Sebastians assumption. Even if it might be the same thread when running in unit-tests (Here we have to write the response to a request manually ans therefore no thread pool can be used), still I don't think that this pattern is invalid. I have seen it when connecting in other code (I didn't invent that pattern). Blocking inside the protocol handler indeed would be bad, but in this case we have to block somewhere, if we want to shut-down gracefully. Another option would be to return the closable future and to wait for that to finish outside of PLC4X, but that would make things needlessly more complex for our users - I think.

JulianFeinauer:
@chrisdutz I guess due to my partial scala background I totally ignore Javas Completable futures as superior and thus didnt think about using them : D
But I agree that it shows the intention of the code clearer.
Perhaps we add a Util Class that "converts" a Netty Future into a Completable Future, as we surely need this functionality more often. What do you think?

chrisdutz:
I think there is no need to really convert between the two and their existence is more historically. Netty preceeds the existence of the Java 8 CompletableFuture by far and by not using them, they have no problems running Netty applications on Java 7 VMs. 

Would it be ok for you to close and drop this PR?

JulianFeinauer:
Of course I can drop the PR (I think the discussion was more viable than these 3 lines of code).
And the Plans than that you add the functionality or how do we "enble" this "wait for the plc to close connection" feature?

chrisdutz:
Well we don't have to enable anything ... it's now on per default. 
I am using the CompletableFuture for detecting the close and if this doesn't come within the given timeframe, we do it ourselves. This exactly how the Spec expects us to do it :-)

JulianFeinauer:
Okay, perfect, than I'm fine.

PR #11, opened 2018-08-10
Implement basic example of connecting an S7 device to Azure IoT Hub

DerDackel:
Just a minimal PoC to deliver some bytes read from an S7 controller to Azure IoT Hub via MQTT

### DIFF
--- a/examples/azure/pom.xml
+++ b/examples/azure/pom.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


chrisdutz:
Hi, thanks for this pull request. 
However could I ask you to rather create a new example in the example section containing only that code?

DerDackel:
Whatever do you mean? ;)

Realized my mistake immediately after submitting this PR and re-pushed the whole branch.

chrisdutz:
Well I had to do some maven magic as there seem to have been contradicting maven versions in the transitive deps, added Apache headers to all files and fixed a little problem in the program (The "OUTPUTS/0" Address was hard-coded ... replaced that with a 3rd argument. 

Thanks for this first contribution to Apache PLC4X. 

We would be very happy to see more commits like this in the future :-)

PR #21, opened 2018-09-17
Made PlcInvalidFieldException.java checked.

JulianFeinauer:
Made PlcInvalidFieldExeption checked, added UnchekedInvalidFieldException and changed everywhere.


JulianFeinauer:
Hi @sruehl I think I cannot answer to your question directly, I think.
I get your point and in these points you point out it's true that I do not know what the appropriate handling is.
I opened a dicussion thread on the list to exchange the arguments and find a common understanding :)

JulianFeinauer:
Hey @chrisdutz and @sruehl just wanted to bump that up. I think we should decide here on something and either merge or decline the PR. What do you hink about the checked exceptions for wrong user inputs?

sruehl:
As I said: The annoyance to be forked to catch a exception would outweigh the benefit. In the usual api use this would a one-time-occur-exception so we would require the developer to pollute the code with catch(e) new RuntimeException(e) snippets. So I would suggest/vote to close this PR and maybe pickup the topic in later releases (if at all).

JulianFeinauer:
Closing as suggested by sebastian.
But I definitly like to adress the topic nonetheless as we heavily depend on user input which can always be errorous.

PR #23, opened 2018-09-18
Added note for Service Loader.

JulianFeinauer:
I'm sorry for this small PR but I wanted to document the result of our discussion (also for myself) and I cant push directly to master : )

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriver.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriver.java
@@ -24,6 +24,10 @@ import org.apache.plc4x.java.api.exceptions.PlcConnectionException;
 
 /**
  * General interface defining the minimal methods required for adding a new type of driver to the PLC4J system.
+ *
+ * <b>Note that each driver has to add a service file called org.apache.plc4x.java.PlcDriver to
+ * src/main/resources/META-INF which contains the fully qualified classname in order to get loaded
+ * by the PlcDriverManager instances.</b>
  */
 public interface PlcDriver {
 
--- a/examples/google/src/main/java/org/apache/plc4x/java/examples/google/iotcore/S7PlcToGoogleIoTCoreSample.java
+++ b/examples/google/src/main/java/org/apache/plc4x/java/examples/google/iotcore/S7PlcToGoogleIoTCoreSample.java
@@ -18,16 +18,19 @@ under the License.
 */
 package org.apache.plc4x.java.examples.google.iotcore;
 
+import io.jsonwebtoken.JwtBuilder;
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.SignatureAlgorithm;
 import org.apache.plc4x.java.PlcDriverManager;
 import org.apache.plc4x.java.api.connection.PlcConnection;
 import org.apache.plc4x.java.api.connection.PlcReader;
 import org.apache.plc4x.java.api.messages.PlcReadRequest;
 import org.apache.plc4x.java.api.messages.PlcReadResponse;
-
-// [START iot_mqtt_includes]
-import io.jsonwebtoken.JwtBuilder;
-import io.jsonwebtoken.Jwts;
-import io.jsonwebtoken.SignatureAlgorithm;
+import org.eclipse.paho.client.mqttv3.*;
+import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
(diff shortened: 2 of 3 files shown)


JulianFeinauer:
@sruehl thanks for the hint, good catch.
I have problems mit mvn site:run and filed jira PLC4X-52.

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriver.java
```diff
@@ -24,6 +24,10 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 /**
  * General interface defining the minimal methods required for adding a new type of driver to the PLC4J system.
+ *
+ * <b>Note that each driver has to add a service file called org.apache.plc4x.java.PlcDriver to
+ * src/main/resources/META-INF which contains the fully qualified classname in order to get loaded
+ * by the {@link java.sql.DriverManager} instances.</b
```
sruehl: `</b` is missing the closing `>`

sruehl: that might be where the first has gone it the first place :)

JulianFeinauer:
Oh my... that was a classical "i fix this in a second and then...".
Sorry for that and thanks for the catch. I will ask chris on the list to look into the maven problems (if he doesnt already) to be able to run "tests"

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriver.java
```diff
@@ -24,6 +24,10 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 /**
  * General interface defining the minimal methods required for adding a new type of driver to the PLC4J system.
+ *
+ * <b>Note that each driver has to add a service file called org.apache.plc4x.java.PlcDriver to
+ * src/main/resources/META-INF which contains the fully qualified classname in order to get loaded
+ * by the {@link java.sql.DriverManager} instances.</b>
```
sruehl: `java.sql.DriverManager` shouldn't that be the `org.apache.plc4x.java.PlcDriverManager`?

JulianFeinauer:
@sruehl I noticed that I wasnt able to reference the PlcDriverManager as it is placed in core. So I suggest to extract an interface PlcDriverManager and to rename the implementation to DefaultPlcDriverManager.
The Interface is especially necessary for us to implement our proxy, as the PlcDriverManager can be used as an equivalent of the javax.sql Datasource.

sruehl:
> @sruehl I noticed that I wasnt able to reference the PlcDriverManager as it is placed in core. So I suggest to extract an interface PlcDriverManager and to rename the implementation to DefaultPlcDriverManager.
> The Interface is especially necessary for us to implement our proxy, as the PlcDriverManager can be used as an equivalent of the javax.sql Datasource.

fair enough im just thinking if there would be a better way to retrieve the PlcDriverManager as to use the `DefaultPlcDriverManager`. Might confuse a Api User but im not sure what would be best though...

JulianFeinauer:
I agree, and I also did not like that.
I shortly thought about addind a static factory method to the interface but this is something that I dont like that much.
I'm not sure what the best way is as it feels right to have the interface in the api module but I agree that it is confusing for users.

sruehl:
What we could do I add a Interface on top of the existing driver manager like `IPlcDriverManager` or something and write in the jdoc that it needed for proxies. This way the usage stays the same and the purpose is clear. But dunno :/

JulianFeinauer:
I'm also unsure about that.
We can also leave it as is but perhaps we should move the PlcDriverManager to the api package?

Another approach would be to use a PlcDriverManagerFactory witha  static create method (kind of like the entity manager).

chrisdutz:
I would have left the DriverManager (But I do like the name PlcDriverManager more) in the API module, but the problem was that at that time I couldn't write any tests for that as that would have required a cyclic dependency in the maven project graph. I definitely agree that it would be cooler to just depend on the API module. Even if the DriverManager isn't really an API class ... but I guess users would probably forgive us that.

sruehl:
the factory `PlcDriverManagerFactory` approach might be the best at the end... but keep in mind that this factory can't the reference the default one directly as we would get an circular dependency

JulianFeinauer:
Yes... this is kind of tricky.
A kingdom for some injection frameworks.. : )
I suggest to revert my changes and remove the link in the javadoc to make this merge-able.
Then we discuss the topic seperate, what do you think?

sruehl:
good, idea. Maybe you could squash/soft-reset your changes into one commit.

JulianFeinauer:
@chrisdutz So you suggest to move the PlcDriverManager to the api module to resolve the problem?

sruehl:
> @chrisdutz So you suggest to move the PlcDriverManager to the api module to resolve the problem?

as far as I Understand we wouldn't have the cyclic dependency problem then and you could reference the default one from a factory and still use the interface for your proxy...

chrisdutz:
I think things changed with the latest refactoring of the API as we no longer have real classes, just interfaces ... I guess we could try moving that class there and get rid of the "core" module. Would definitely be worth it. I just knew that I wanted it in the API but I had to introduce the core in order to resolve the cyclic dependencies (At that time ... might be obsolete now)

PR #24, opened 2018-09-26
WIP: Feature/top level item spliting

sruehl:
generic top item level splitting which splits a batch request into multiple requests so protocols can use this layer where batching isn't possible at all.


chrisdutz:
So do I get this correctly, that with the SingleItemToSingleRequestProtocol is a simplified and distilled version of what I did in the S7 protocol?

sruehl:
in theory we could just insert it into ADS and modbus and they could support unoptimized multiple fields till that is implemented. But this all needs testing though. The PR is just to keep this work visible as the branch was several months behind.

PR #25, opened 2018-09-27
PLC4X-57 Bugfix

JulianFeinauer:
Fixed PLC4X-57.
Also fixed Output from HelloPlc4x.java.
Fixed some small bugs.
Added TODO for returning arrays on bit streams and chars / strings.

### DIFF
--- a/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/netty/Plc4XS7Protocol.java
+++ b/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/netty/Plc4XS7Protocol.java
@@ -21,12 +21,10 @@ package org.apache.plc4x.java.s7.netty;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
-import org.apache.plc4x.java.api.exceptions.PlcException;
-import org.apache.plc4x.java.api.exceptions.PlcIoException;
-import org.apache.plc4x.java.api.exceptions.PlcProtocolException;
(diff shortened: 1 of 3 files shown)


JulianFeinauer:
Sorry, the rebase did not work and I had to do it in a second commit.
@chrisdutz I added the functionality to return arrays in the Plx4XS7Protocol.java and it would be good if you look over it and my todo there.
Be aware, the code is pretty ugly due to some Type Erasure problems (Generic Arrays), especially in the bit-string case (but it was ugly there before so ... :D).

plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/netty/Plc4XS7Protocol.java
```diff
@@ -420,22 +423,29 @@ private PlcResponse decodeReadResponse(S7ResponseMessage responseMessage, PlcReq
...
+                // TODO 2018-09-27 jf: array returning only implemented for BOOL, BYTE, INTEGERS, FP
+                // not for CHARS & STRINGS and not for all other bit-strings except for BYTE
                 switch (field.getDataType()) {
                     // -----------------------------------------
                     // Bit
                     // -----------------------------------------
                     case BOOL: {
-                        byte byteValue = data.readByte();
-                        fieldItem = new S7BooleanFieldItem(field.getDataType(),byteValue != 0x00);
+                        Boolean[] booleans = readAllValues(field, i -> data.readByte() != 0x00).toArray(new Boolean[field.getNumElements()]);
```
sruehl: actually sonar finds a presized array as a bug: see here https://stackoverflow.com/a/29444594/850036

JulianFeinauer: I'm about to fix it... didn't know that

chrisdutz:
By the way ... looking good ... just had a few minutes after lunch and checked out your changes. 
And I already added a test for the field with array with a more than one digit number of elements in a separate test case.

PR #26, opened 2018-09-30
connection manager for connection pool initial version

Compater:
As discussed at the meet up this is the first implementation of our connection manager acting as a connection pool.

### DIFF
--- a/plc4j/core/src/main/java/org/apache/plc4x/java/PlcDriverManagerInterface.java
+++ b/plc4j/core/src/main/java/org/apache/plc4x/java/PlcDriverManagerInterface.java
@@ -1,0 +1,29 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+package org.apache.plc4x.java;
+
+import org.apache.plc4x.java.api.authentication.PlcAuthentication;
+import org.apache.plc4x.java.api.connection.PlcConnection;
+import org.apache.plc4x.java.api.exceptions.PlcConnectionException;
+
+public interface PlcDriverManagerInterface {
+    PlcConnection getConnection(String url) throws PlcConnectionException;
+
+    PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException;
+}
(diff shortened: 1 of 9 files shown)


plc4j/core/src/main/java/org/apache/plc4x/java/PlcDriverManagerInterface.java
```diff
@@ -0,0 +1,29 @@
...
+specific language governing permissions and limitations
+under the License.
+*/
+package org.apache.plc4x.java;
+
+import org.apache.plc4x.java.api.authentication.PlcAuthentication;
+import org.apache.plc4x.java.api.connection.PlcConnection;
+import org.apache.plc4x.java.api.exceptions.PlcConnectionException;
+
+public interface PlcDriverManagerInterface {
```
JulianFeinauer: I dont like the name that much, but I see your problem (as we want to keep the name of the "original" PlcDriverMangager). Perhaps it is best to simply subclass the existing PlcDriverManager as we want to allow users to use the Proxy AS PlcDriverManager. So it should be:
`PlcDriverManager manager = new ...`

plc4j/core/src/main/java/org/apache/plc4x/java/connectionpool/ConnectionProxy.java
```diff
@@ -0,0 +1,24 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+package org.apache.plc4x.java.connectionpool;
+
+import org.apache.plc4x.java.api.connection.PlcConnection;
+
+public interface ConnectionProxy extends PlcConnection {
```
JulianFeinauer: Do we need this? I guess not?

plc4j/core/src/main/java/org/apache/plc4x/java/connectionpool/PlcConnectionManager.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.annotation.PreDestroy;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Plc Connection Manager to act as a controller to give threads access to a connection in the pool only if it is not
+ * used by a different thread
+ */
+public class PlcConnectionManager implements PlcDriverManagerInterface {
```
JulianFeinauer: Perhaps it is more intuitive to name the Thing PlcConnectionPool or something.
And it should (as stated above) implement PlcDriverManager directly.

plc4j/core/src/main/java/org/apache/plc4x/java/connectionpool/PlcConnectionManager.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(PlcConnectionManager.class);
+    //TODO: implement connection Pool with more than one connection
+    private Map<String, PlcConnection> connections = new HashMap<>();
```
JulianFeinauer: I think we have to use a ConcurrentMap or to use locks to avoid ConcurrentModificationExceptions, as multiple threads could modify the list concurrently.

JulianFeinauer:
This PR *should* not be merged but all necessary changes should be done in #30.

sruehl:
super seeded by commons-pool

PR #28, opened 2018-10-06
Type conversions for default byte array fiel item

timbo2k:
implemented conversion methods for
* getShort
* getInteger
* getLong

as fix for PLC4X-62


sruehl:
@timbo2k @JulianFeinauer initialy I left out this methods on purpose as you can't imply endianess of the data. Otherwise one could use the the DefaultLongFieldItem etc. in the first place. What could be a sollution is to add the getLongBE() / getLongLE() methods on the DefaultByteArrayFieldItem and do a instanceOf check / downcast in the client.

JulianFeinauer:
@sruehl Hey seb, you mean the Ednainess could change? Or is undefined? Then I would suggest to add a connection parameter (as discussed some times) like modbus://...?endianess=big or bigEndian=true wiht a sensible default.

sruehl:
@JulianFeinauer modbus itself doesn't say anything about the payload itself. That might work, however one restriction would be that then all data is read in the same endinaness from this particular plc and if you need other endianess you would need a second connection. This might be a acceptable drawback (haven't thought this through yet)...

JulianFeinauer:
@sruehl so in theory could someone crazy enough decide to have different memory areas in its plc in different endianess??? Well... I'm not sure if I want to support communication with those people :)
But back to topic, I would suggest to do it in the connection or in the request query (similar to what we do in S7 like "/0/0:LE" and have a default like BE then we are flexible enough.
What I want to avoid (and I'm pretty sure @chrisdutz is with me in this regard is to have the users think about this stuff like do a 
`if (x.isBE) {
 return x.getLongBE()
}`
This looks horrible.

sruehl:
@JulianFeinauer pretty sure there are some out there 😛 
Im fine with both suggestions. :)

JulianFeinauer:
So @timbo2k , @sruehl what about the suggestion:
We copy the DefaultByte... (as commited by Tim) in the Modbus module and name is something like DefaultModbusItem.
Parallel we do a Improvement Ticket to add a syntax somewhere to speciy Little Endianess (and Big is the default).

sruehl:
good Idea (this could extend from the DefaultByteArrayItem), and at best add a TODO and JIRA-ISSUE id to the BIG_ENDIAN call in the field item :)

timbo2k:
@JulianFeinauer @sruehl that sounds like a good idea ... 
I didn't thought about endianess but u'r totally right.
I'll change those stuff and update PR

timbo2k:
PR has been updated ... explicit Modbus-Implementation for ByteArrayFieldItem makes sense
Code change has been successfully tested vs an industrial Modbus device

chrisdutz:
Regarding the endianess of data ... I guess every PLC is either Little Endian or Big Endian. How about providing this detail as part of the connection string? I think the Client code shouldn't depend on the endianess of the PLC.

chrisdutz:
An I should have read through the entire thread here first ;-) ... So I am on the same page as Julian. We should set the endianess on a connection level. I haven't come across any PLC where the types differed. Regarding the connection string ... I would make it mandatory:
modbus://10.10.40.10:234/LE or modbus://10.10.40.10:234/BE ... it is an important detail similar like "Rack" and "Slot" for S7.

JulianFeinauer:
PR is closed an I opened a JIRA (https://issues.apache.org/jira/projects/PLC4X/issues/PLC4X-63?filter=allopenissues) for adding endianess to the connection string.

PR #29, opened 2018-10-22
Opm plcentitymanager

JulianFeinauer:
I just finished the first (working) implementation of the Object-Plc-Mapping (OPM).
This is a JPA Like functionality on top of plc4x which allows to issue requests against plc4x by calling Methods on POJOs!

Basically, it uses 3 classes.

  *   PlcEntityManager: This is the main class. A call on read(…) or connect(…) does all the necessary magic
  *   PlcEntity (Annotation): A Pojo needs this annotation to be used with read(…) or connect(…) above
  *   PlcField (Annotation): Fields in the Pojo that are annotated with this field are fetched from the plc

Some simple tests (with mocking) can be found in PlcEntityManagerTest.

I would be really happy for some feedback on the basic semantics (which annotations, where to put them, …) or even some testing.
If the feedback on this is positive I would like issue a PR for this.


plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntity.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a class as plc entity.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface PlcEntity {
+    String value();
```
sruehl: maybe we can add a additional String connectionUrl() method that is semantical the same as value() but haven't found a elegant way to achieve this.

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+        this.driverManager = driverManager;
+    }
+
+    public <T> T read(Class<T> clazz) throws OPMException {
+        PlcEntity annotation = getPlcEntityAndCheckPreconditions(clazz);
+        String source = annotation.value();
+
+        try (PlcConnection connection = driverManager.getConnection(source)) {
+
+            if (!connection.getReader().isPresent()) {
```
sruehl: could be replaced with orThrow()

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+                throw new OPMException("Unable to get Reader for connection with url '" + source + "'");
+            }
+
+            PlcReader reader = connection.getReader().get();
+
+            PlcReadRequest.Builder requestBuilder = reader.readRequestBuilder();
+
+            // Do the necessary queries for all fields
+            // HashMap<ReadRequestItem<?>, Field> requestItems = new HashMap<>();
+            for (Field field : clazz.getDeclaredFields()) {
```
sruehl: later on we might support method injection..

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+            try {
+                request = requestBuilder.build();
+            } catch (PlcInvalidFieldException e) {
+                throw new OPMException("Unable to parse one field request", e);
+            }
+
+            // Perform the request
+            PlcReadResponse<?> response;
+            try {
+                // TODO: make configurable.
```
sruehl: added a todo

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * Returns a connected proxy.
+     *
+     * @param clazz clazz to be connected.
+     * @param <T>   type of param {@code clazz}.
+     * @return a connected entity.
+     * @throws OPMException when proxy can't be build.
+     */
+    public <T> T connect(Class<T> clazz) throws OPMException {
```
sruehl: confuses me a bit. Is this the same as merge in jpa?

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+     * @param that Reference to the proxy object
+     * @return possible result of the original methods invocation
+     * @throws OPMException Problems with plc / proxying
+     */
+    @RuntimeType
+    // TODO: avoid using single letter params.
+    public Object intercept(@This Object o, @Origin Method m, @SuperCall Callable<?> c, @Super Object that) throws OPMException {
+        LOGGER.trace("Invoked method {} on connected PlcEntity {}", m.getName(), that);
+
+        if (m.getName().startsWith("get")) {
```
sruehl: maybe we should assert that no method starting with get and having attributes get picked up

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+    private void refetchAllFields(Object o) throws OPMException {
+        Class<?> superclass = o.getClass().getSuperclass();
+        PlcEntity plcEntity = superclass.getAnnotation(PlcEntity.class);
+
+        try (PlcConnection connection = driverManager.getConnection(plcEntity.value())) {
+            // Catch the exception, if no reader present (see below)
+            PlcReader plcReader = connection.getReader().get();
+
+            // Build the query
+            PlcReadRequest.Builder builder = plcReader.readRequestBuilder();
```
sruehl: so currently only read is possible?

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+                } catch (NoSuchFieldException | IllegalAccessException e) {
+                    throw new PlcRuntimeException(e);
+                }
+            }
+        } catch (Exception e) {
+            throw new OPMException("Problem during processing", e);
+        }
+    }
+
+    // TODO: why isn't o used?
```
sruehl: added a TODO

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
```diff
@@ -0,0 +1,380 @@
...
+     * @throws NoSuchFieldException
+     * @throws IllegalAccessException
+     */
+    private void setField(Class<?> clazz, Object o, PlcReadResponse<?> response, String fieldName) throws NoSuchFieldException, IllegalAccessException {
+        Field field = clazz.getDeclaredField(fieldName);
+        field.setAccessible(true);
+        try {
+            field.set(o, getTyped(field.getType(), response, fieldName));
+        } catch (ClassCastException e) {
+            // TODO should we simply fail here?
```
sruehl: think so as this can be checked beforehand and so an simple fail is sufficient here.

sruehl:
looks good to me :)

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
JulianFeinauer: Its kinda similar to `find(...)` in JPA because it returns a connected entity.
But as it does not have a primary key (makes no sense) I decided to name it connect.
But thinking about it, it is pretty similar to `merge(...)` with the difference that it takes a class and not an instance.
But I'm open for changes in the naming.

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
JulianFeinauer: This is a fair point. And in fact, we should also support `isXXX()` for boolean fields.
We should add a Todo here, or better extract this in a method `isGetter(m)` where we can adopt all these conditions.

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
JulianFeinauer: Yes, perhaps I didnt state this enough. Currently, only read. It should not be that big of a deal to add writing. I'll open a Jira for that.

JulianFeinauer: Done.. is PLC4X-70

plc4j/utils/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityManager.java
JulianFeinauer: In theory this should be an IllegalStateException, as this should never happen, or?

JulianFeinauer:
Thank you @sruehl for the comments.
I answered your comments although it is already merged. I would love to test this feature (together with your pool).

PR #36, opened 2018-10-31
Add simple mock driver

JulianFeinauer:
Hi all,
I wrote a simple Mock Driver implementation, as I need to do some mocking for our application code.
To avoid conflicts I renamed the prefix of the current existing mock driver (in core) to "spi-mock".
It would be good if someone checks over my changes to see if my changes are okay.

Thanks!
Julian


chrisdutz:
I thought it would make more sense to start removing all of these mock drivers and instead extend the "Test-Driver" as this should provide everything that's needed.

On the one side, we can use this internally for testing our integrations and examples.

On the other side users can play around with PLC4X without having a real PLC.

JulianFeinauer:
@chrisdutz I'll check if I can add this Mock functionality in there. As I recall the test driver is currently working against internal values or random values. But what I want is an object to put assertions on, so I'll check that and close this PR for the time beeing.

anup-km:
Hi @JulianFeinauer , I got some PLC4X connection error while running .jar file in command prompt, but when I run the same source code on the IntelliJ environment its getting executing properly and working properly. I gone through the error by debugging and got this "org.apache.plc4x.java.spi.transport.Transport" for this interface the IntelliJ is providing the .class implementation. How to run on the command prompt, please help me out.
Screenshot (17)

anup-km:
> Hi @JulianFeinauer , I got some PLC4X connection error while running .jar file in command prompt, but when I run the same source code on the IntelliJ environment its getting executing properly and working properly. I gone through the error by debugging and got this "org.apache.plc4x.java.spi.transport.Transport" for this interface the IntelliJ is providing the .class implementation. How to run on the command prompt, please help me out.
> Screenshot (17)

Hi any update on this..

PR #37, opened 2018-11-01
Extend test driver with mock

JulianFeinauer:
@chrisdutz here is my approach as extension to the Test Driver.
@sruehl It is still not in the Test-utils package but I'm unsure if it should as I'm unsure about all the dependency mambo jambo.


JulianFeinauer:
Not needed. In PR #38 similar functionality is implemented.

PR #38, opened 2018-11-02
Add simple mock driver

JulianFeinauer:
I decided to make a new PR (in concurrence to #37 ) as this is the solution we need to test our code.
The idea behind this mock driver is to allow testing **WITHOUT** changing the field adresses in your code you want to test (currently they have to be hardcoded for some reasons).
This is not possible with the test driver and so I suggest to merge this pr and close #37 if you agree.
@chrisdutz would be good to have your feedback on that (I know there are many mock drivers.. but I think there is nothing (yet) to solve this problem described above).

PS.: kittens


JulianFeinauer:
Got green light from @chrisdutz and will merge now 👍

PR #42, opened 2018-11-26
[PLC4X-75] - Fixing dependency to the wrap url-handler

ANierbeck:


sruehl:
@ANierbeck thanks for the PR :)

PR #44, opened 2019-01-16
S7: changed byteLength and blockNumber from short to int

timbo2k:
byteLength and blockNumbers had been to small using short for possible values
added some tests for out of range values
did successful manual test with scraper and updated commited yml-config

### DIFF
--- a/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/model/S7Field.java
+++ b/plc4j/protocols/s7/src/main/java/org/apache/plc4x/java/s7/model/S7Field.java
@@ -28,10 +28,13 @@ import java.util.regex.Pattern;
 
 public class S7Field implements PlcField {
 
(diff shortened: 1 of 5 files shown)


chrisdutz:
Argh ... dammit ... you are right. The protocol only transfers 2 bytes (1 short). However you are correct that Java shorts are signed and the protocol uses unsigned, so you are correct that we should change this to int ... however we now have to check if the protocol code is correct everywhere.

chrisdutz:
Ok .. so I double-checked the Protocol implementaton ... on the writing side, we're safe as "writeShort(int)" works like a charm and we currently don't parse "AnyItems", so no way we could parse it wrong. Guess we can merge this.

Thanks for contributing :-)

👍

timbo2k:
I just ran a test vs our physical device with had been successful - so everything seems to be Ok

PR #56, opened 2019-03-28
Feature/connection manager

gg587998:
Hi this is the first version of the connection-manager i wrote for our project. I hope some of the code is usefull to integrate.


gg587998:
in some cases this does not work, i did find the problem yet

PR #57, opened 2019-03-29
Feature/plc4 x 108 ping method

JulianFeinauer:
I implemented a ping() method based on Gunters code (in fact it is Gunters code) and "redirected" the isConnected method there. 
I extended the NettyPlcConnection and the ConnectionFactory to use the ping based approach *if and only if* the protocol has ip adress and port.
But for all these, it should work out of the box now.

A manual test was performed by @timbo2k and me with S7 and Modbus-TCP and both worked excellent.

### DIFF
--- a/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
+++ b/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
@@ -83,13 +83,14 @@ public abstract class AbstractPlcConnection implements PlcConnection, PlcConnect
      * Otherwise it throws a {@link NotImplementedException} to inform the user about that.
      */
     protected boolean ping(int timeout) {
-        Optional<InetSocketAddress> address = getInetSocketAddress();
-        if (!address.isPresent()) {
-            throw new NotImplementedException("Tries to check the connection with ping, " +
-                "but no Socket Address is given!");
-        }
+        InetSocketAddress address = getInetSocketAddress()
+            .orElseThrow(() -> new NotImplementedException("Tries to check the connection with ping, " +
+                "but no Socket Address is given!"));
+
         try (Socket s = new Socket()) {
-            s.connect(address.get(), timeout);
+            s.connect(address, timeout);
+            // TODO keep the adress for a (timely) next request???
+            s.setReuseAddress(true);
             return true;
         } catch (Exception e) {
             LOGGER.debug("Unable to ping PLC", e);
--- a/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
+++ b/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
@@ -18,6 +18,7 @@ under the License.
 */
 package org.apache.plc4x.java.base.connection;
 
+import org.apache.commons.lang3.NotImplementedException;
 import org.apache.plc4x.java.api.PlcConnection;
 import org.apache.plc4x.java.api.exceptions.PlcUnsupportedOperationException;
 import org.apache.plc4x.java.api.messages.PlcReadRequest;
@@ -26,8 +27,13 @@ import org.apache.plc4x.java.api.messages.PlcUnsubscriptionRequest;
 import org.apache.plc4x.java.api.messages.PlcWriteRequest;
 import org.apache.plc4x.java.api.metadata.PlcConnectionMetadata;
 import org.apache.plc4x.java.base.messages.InternalPlcMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.net.InetSocketAddress;
+import java.net.Socket;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Base class for implementing connections.
@@ -36,6 +42,10 @@ import java.util.Objects;
  * and for obtaining respective request builders.
  */
 public abstract class AbstractPlcConnection implements PlcConnection, PlcConnectionMetadata {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractPlcConnection.class);
+
+    private static final int PING_TIMEOUT_MS = 1_000;
 
     @Override
     public PlcConnectionMetadata getMetadata() {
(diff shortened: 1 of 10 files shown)


plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -57,6 +67,51 @@ public boolean canSubscribe() {
...
+     * Otherwise it throws a {@link NotImplementedException} to inform the user about that.
+     */
+    protected boolean ping(int timeout) {
+        Optional<InetSocketAddress> address = getInetSocketAddress();
+        if (!address.isPresent()) {
+            throw new NotImplementedException("Tries to check the connection with ping, " +
+                "but no Socket Address is given!");
+        }
+        Socket s = null;
+        try {
```
sruehl: may I suggest to use try-with-ressource here as socket is auto closable

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -57,6 +67,51 @@ public boolean canSubscribe() {
...
+                    s.close();
+                } catch (Exception e) {
+                }
+            }
+        }
+    }
+
+    /**
+     * Strategy Pattern method for the {@link #ping(int)} method.
+     * If a driver has no Inet Socket adress, it has to return an Empty Optional, never null.
```
sruehl: the description of optional is a bit redundant here. never returning null is the idea of optional

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
JulianFeinauer: Full-ack, good catch. Just updated the PR.

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
JulianFeinauer: Indeed this is redundandent. But I'm a bit "conservative" with these Java 8 features as there may be people not so used to them so I think its okay to make it *really* clear :)
Better safe than sorry.

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -57,6 +67,42 @@ public boolean canSubscribe() {
...
+    }
+
+    /**
+     * Method that sends a test-request or ping to the PLC to check if the PLC is still there.
+     * In most cases this method should only be used from the {@link #isConnected()} method.
+     * This method can only be used if ghe {@link #getInetSocketAddress()} returns a Socket Adress.
+     * Otherwise it throws a {@link NotImplementedException} to inform the user about that.
+     */
+    protected boolean ping(int timeout) {
+        Optional<InetSocketAddress> address = getInetSocketAddress();
```
sruehl: You could write this in one line with orElseThrow()

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -57,6 +67,42 @@ public boolean canSubscribe() {
         return false;
     }
 
+    /**
+     * The default implementation uses the {@link #ping(int)} method.
+     * Drivers that want to implement a more specific version have to overide it.
+     */
+    @Override
+    public boolean isConnected() {
+        return ping(PING_TIMEOUT_MS);
```
sruehl: could we prepend a `channel.isActive()` here as well?

sruehl: with a or so we don't try to send a ping in first place if we don't have an active channel

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -57,6 +67,42 @@ public boolean canSubscribe() {
...
+     * Otherwise it throws a {@link NotImplementedException} to inform the user about that.
+     */
+    protected boolean ping(int timeout) {
+        Optional<InetSocketAddress> address = getInetSocketAddress();
+        if (!address.isPresent()) {
+            throw new NotImplementedException("Tries to check the connection with ping, " +
+                "but no Socket Address is given!");
+        }
+        try (Socket s = new Socket()) {
+            s.connect(address.get(), timeout);
```
sruehl: I'm not sure if we should call .disconnect() here as well. Does socket.close() does this. If thats isn't the case we should call it so the socket doesn't stay open on the other side

sruehl: and the disconnect would hurt. What we could do though is using SO_REUSEADDR (if that applies for clients at all) to speed things up

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
JulianFeinauer: done.

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
JulianFeinauer: will add a todo, as I'm unsure if we should do that or not...

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
JulianFeinauer: Just checked the docs (https://docs.oracle.com/javase/7/docs/technotes/guides/net/articles/connection_release.html) and it seems that #close() does all thats necessary.
I added the SO_REUSEADDR but I'm unsure ablout it. Maybe others have an opinion about that?

sruehl: might only apply to ServerSockets so we should be good here: https://docs.oracle.com/javase/7/docs/api/java/net/StandardSocketOptions.html#SO_REUSEADDR

PR #59, opened 2019-04-10
Feature/alternate plc4 x 108

chrisdutz:
Hi,

in this alternate implementation, I added a "ping" method to the default API and moved the code for the TCP ping into the classes actually involved in TCP (with options for other base types - serial, udp, raw-socket, test, ...).

Please review.

### DIFF
--- a/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
+++ b/plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
@@ -18,7 +18,6 @@ under the License.
 */
 package org.apache.plc4x.java.base.connection;
 
-import org.apache.commons.lang3.NotImplementedException;
 import org.apache.plc4x.java.api.PlcConnection;
 import org.apache.plc4x.java.api.exceptions.PlcUnsupportedOperationException;
 import org.apache.plc4x.java.api.messages.PlcReadRequest;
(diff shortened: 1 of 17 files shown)


JulianFeinauer:
@chrisdutz thank you for the PR! I will try to review it this week and start some tests with our S7 PLC here, to see if it works.

plc4j/protocols/driver-bases/base/src/main/java/org/apache/plc4x/java/base/connection/AbstractPlcConnection.java
```diff
@@ -43,15 +38,18 @@ Licensed to the Apache Software Foundation (ASF) under one
...
-
     @Override
     public PlcConnectionMetadata getMetadata() {
         return this;
     }
 
+    @Override
+    public CompletableFuture<Void> ping() {
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        future.completeExceptionally(new PlcUnsupportedOperationException("The connection does not support pinging"));
```
sruehl: I would suggest to use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html#failedFuture(java.lang.Throwable) here if possible.

chrisdutz: Argh ... was looking for something like that ... strangely I must have missed it :-) ... of course ... will update that.

chrisdutz: Ha! ... I knew there was something ... if I used the failedFuture method, we would have to bump up the Java level to 9 which I wasn't planning on doing just for this one convenience method.

sruehl: fair enough. :)

PR #67, opened 2019-05-14
poll loop fetches data in a continous loop

gg587998:
Hi Christofer, sorry for the late reply.  I created a new pull request.


chrisdutz:
Hi,

could you please elaborate a bit about what this example does?
And sorry for not addressing this for so long ... has been quite busy lately ;-)

Chris

gg587998:
the example is a mechanism to continously fetch data from a S7. I and maybe a lot of other people have the following use case: we have some variables we want to read from the s7 and these are monitored, visualized e.g. so we fetch the same variables every second and output them.
* the code automatically reconnects when the connection is lost 
* there is sample code how to manually initialize the s7 driver in a osgi environment, where the mechanism of automatically loading the driver does not work

chrisdutz:
Oh gee .. I can't believe how long this has been open .. I am extremely sorry for that. 
I just had a look and to me it seems as if the same thing you are addressing here is also being done by the "scraper" component in PLC4X ... I just saw that it's still 100% undocumented on the website, but you can see in the Logstash and Kafka Connect adapters on how to use it.

chrisdutz:
Guess it doesn't harm adding this example ... perhaps it helps someone ... however I am not sure if we would need an ICLA on file for this contribution.

PR #71, opened 2019-05-27
PLC4X-129 Improve version detection macro.

splatch:
Attached PR contains changed regexp. Additionally same regexp is used for all programs as it was copy/pasted all over the place.


JulianFeinauer:
🎉 Congratz to your first MERGED PR Lukasz 👍

PR #75, opened 2019-08-01
fix bug in opc ua driver response handling when fields are not found

stefah:
Fixes a bug within the check for returned fields in the opc ua driver connection response, where the status code was wrongly compared.
Updates the milo dependencies.

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
@@ -346,7 +346,7 @@ public class OpcuaTcpPlcConnection extends BaseOpcuaPlcConnection {
             for(int counter = 0; counter < readValueIds.size(); counter++){
                 PlcResponseCode resultCode = PlcResponseCode.OK;
                 BaseDefaultFieldItem stringItem = null;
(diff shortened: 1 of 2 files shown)


chrisdutz:
Looks good to me ... and I did see it fail before and work with the change ;-)

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
```diff
@@ -346,7 +346,7 @@ public void unregister(PlcConsumerRegistration registration) {
             for(int counter = 0; counter < readValueIds.size(); counter++){
                 PlcResponseCode resultCode = PlcResponseCode.OK;
                 BaseDefaultFieldItem stringItem = null;
-                if(readValues == null || readValues.size() <= counter || readValues.get(counter).getStatusCode() != StatusCode.GOOD){
+                if(readValues == null || counter > readValues.size() || !readValues.get(counter).getStatusCode().isGood()){
```
nalim2: Correct but it has to be `counter >= readValues.size()` or?

chrisdutz: Actually I think you are right as if it's equal to size it must produce an array index out of bounds exception. Stefan, could you please re evaluate that?

stefah: you are correct, will fix that.. :)

PR #76, opened 2019-08-04
PLC4X-139 close the worker thread on connection abortion to avoid thr…

JulianFeinauer:
…ead and socket leak.

I would like two people to review this change.
I am not able to write a test.
All I do is to start multiple connections (that fail, see https://issues.apache.org/jira/browse/PLC4X-139?jql=project%20%3D%20PLC4X) and then check how the number of sockets behaves (also see my description in the PR).

For my tests it stabilizes around 3k when I do 20 connections in parallel (which all fail).
Without the fix it goes to around 10k and then "too many open files" exceptions start to occur.

### DIFF
--- a/plc4j/protocols/driver-bases/tcp/src/main/java/org/apache/plc4x/java/base/connection/TcpSocketChannelFactory.java
+++ b/plc4j/protocols/driver-bases/tcp/src/main/java/org/apache/plc4x/java/base/connection/TcpSocketChannelFactory.java
@@ -25,14 +25,20 @@ import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
 import org.apache.plc4x.java.api.exceptions.PlcConnectionException;
 import org.apache.plc4x.java.api.exceptions.PlcException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 
 public class TcpSocketChannelFactory implements ChannelFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger(TcpSocketChannelFactory.class);
 
     private static final int PING_TIMEOUT_MS = 1_000;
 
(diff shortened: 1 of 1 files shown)


timbo2k:
Hi Julian

i tested your implementation in master and in your branch and watched open files with lsof as u did.
it looks like your solution fixes the problem.
I would like to test the issue vs a real device that loses connection due to Firewall topics or by physical  problem - network broken, lack of power. 

In general to avoid such kind of errors it will be great if can define some kind of tests that check those conditions.

thanks for your work!

plc4j/protocols/driver-bases/tcp/src/main/java/org/apache/plc4x/java/base/connection/TcpSocketChannelFactory.java
```diff
@@ -49,14 +55,27 @@ public Channel createChannel(ChannelHandler channelHandler)
...
             bootstrap.handler(channelHandler);
             // Start the client.
-            ChannelFuture f = bootstrap.connect(address, port).sync();
-            f.awaitUninterruptibly();
+            final ChannelFuture f = bootstrap.connect(address, port);
+            f.addListener(new GenericFutureListener<Future<? super Void>>() {
+                @Override public void operationComplete(Future<? super Void> future) throws Exception {
+                    if (!future.isSuccess()) {
+                        logger.info("Unable to connect, shutting down worker thread.");
+                        workerGroup.shutdownGracefully();
```
niclash: shutdownGracefully() needs the pre-requistite that new tasks are not being added, otherwise the "quiet period" restarts when one is added (according to Netty doc https://netty.io/4.1/api/io/netty/util/concurrent/EventExecutorGroup.html). I am not familiar enough with the details around this, but I thought I bring your attention to such detail. Cheers

JulianFeinauer: Hey @niclash thanks for the input. It shouldnt be a problem in this situation as this is the situation that connection fails so no work item is ever added but before this fix, the pool kept alive forever.
So I think the prerequisite is given here, from my understanding, what do you think, Nick?

niclash: @JulianFeinauer I am sure you know how things are working. I just had a look and recognized a potential issue, one that I had way back in time and caused me a lot of grief. :-)

JulianFeinauer: Haha, yeah, these thinks are really.. nasty. But thanks both of you for your comments!

PR #77, opened 2019-08-08
Feature/improve scraper tim

JulianFeinauer:
@timbo2k last chance : )


timbo2k:
checked my stuff quite detailed and found some bugs ... build is also now green so i'll merge that

PR #81, opened 2019-09-02
Simplifying inheritance of shade plugin configurations

rvs:
It seems that a slightly simpler version of achieving the same behavior as is currently in place is to rely on default Maven inheritance mechanisms to get `transformer implementation=...` configured.

This, of course, requires `combine.children="append"` to be added back at the examples pom.

Or to put it another way, this tries to be a simpler version of:
    https://github.com/apache/plc4x/commit/97e1eacfa912dbc106c955867b57041b58c34ae2

### DIFF
--- a/plc4j/examples/pom.xml
+++ b/plc4j/examples/pom.xml
@@ -74,8 +74,7 @@
               <goal>shade</goal>
             </goals>
             <configuration>
-              <transformers>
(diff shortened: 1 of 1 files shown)


chrisdutz:
Yeah ... have to admit I was just to lazy to find out where to put it ;-) ... thanks for this.

PR #82, opened 2019-09-04
Feature/implement logstash integration

cc-till:
- documentation
- removed some TODOs

Please review the source code regarding scraper implementation (sleep the thread - otherwise the plugin terminates and logstash will shutdown):
```
while(scraper.isRunning()) {
            try {
                Thread.sleep(1000);
                // or maybe just yield(); ?
            } catch (InterruptedException e) {
                logger.severe("Error thead sleep plc4x plugin: "+ e);
            }
        }
```


chrisdutz:
Looks good to me ... think perhaps Julian or Tim can help with your question if the while loop is the ideal solution.

PR #87, opened 2019-10-08
Adding OSGi processing to the pom for java modules

cschneider:


JulianFeinauer:
Thank you... I will have a look when I’m back at the desk

plc4j/pom.xml
```diff
@@ -45,7 +45,37 @@
 
   <build>
     <plugins>
-
+        <plugin>
```
JulianFeinauer: Hey @cschneider there is another invocation of the maven-bundle-plugin (L:200). Is this okay or hould this be removed?

JulianFeinauer: And I would also add the same part (with the part from L:200 with maven-bdunle-plugin and jar plugin???) to protocols/pom.xml, right?

cschneider: Oh .. sorry I missed that it was already present. So yes the other invocation should be removed. Actually to make the diff look cleaner I will merge my changes into the existing config.

cschneider: It also makes sense to add the same bundling config to protocol. I can update the PR.

JulianFeinauer: Thank you : )
Then I will check how the bundles look like and if some of them already work (one can hope...).

JulianFeinauer:
Looks good and builds locally, will merge. Congrats @cschneider for your first merged PLC4X PR 🎉 !

PR #89, opened 2019-10-13
Feature/resolve split package osgi

JulianFeinauer:
This PR brings full OSGi support to PLC4X.
A bundle for the S7 driver is also added.


JulianFeinauer:
Its not intended to be merged, yet its. First, we should check it in-depth.

plc4j/drivers/s7/pom.xml
```diff
@@ -118,6 +118,12 @@
       <version>1.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
chrisdutz: I think this could eventually cause problems. By setting the scope to provided, it is not shipped with the application. So as soon as you run it in a normal environment without osgi, it could cause CNFEs when loading the class which is annotated with "@Component". Will test this.

JulianFeinauer: @chrisdutz I see your point BUT the annotation is not kept in the byte code, its scope is source only. So it will no longer be available when shipped.
(The plugin creates a xml file which is shipped with the code under OSGI-INF/...)

JulianFeinauer:
And I agree with you @chrisdutz with the naming. I just wanted to make the smalles changes possible which was base -> xxx.

ottobackwards:
How about throwing in a readme in the karaf features area? 
I would have expected to see some documentation around OSGI support as well at the top level somewhere

chrisdutz:
We already had "plc4j/integrations/apache-karaf" modules and there we were using the goal "features-generate-descriptor" which generates the "feature.xml" ... the way you're currently doing it would require to manually maintain a second resource with dependency information.

chrisdutz:
Ok ... I compared the feature.xml you added manually, that's identical to the one generated by the plugin, except that it has one additional entry:     <feature>scr</feature>

plc4j/karaf-features/karaf-itest/src/test/resources/etc/system.properties
```diff
@@ -0,0 +1,18 @@
+###############################################################################
```
jbonofre: You don't need this file, it will be created by `editConfigurationPut()` method using the Karaf original one (from the archive).

plc4j/karaf-features/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <targetPath>${project.build.directory}/feature</targetPath>
+      </resource>
+    </resources>
+
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.karaf.tooling</groupId>
+          <artifactId>karaf-maven-plugin</artifactId>
+          <version>4.2.4</version>
```
jbonofre: Why not using 4.2.7 there ?

JulianFeinauer: Good comment... resolved now.

jbonofre: Cool thanks ! Let me do a new pass.

PR #91, opened 2019-10-16
PLC4X-144 - When requesting invalid addresses, the DefaultS7MessagePr…

chrisdutz:
I was able to reproduce the error when talking to a real machine and pretty quickly found the problem. However I spent 1,5 hours trying to whip up a test to reproduce the error and failed to do so.


JulianFeinauer:
@chrisdutz I cannot judge on the exact changes, it looks "ok". Did you run some tests against a PLC of yours (Logo?)? Otherwise I would try to do this later today or tomorrow...

chrisdutz:
I tested it against the S7-1200 ... the logo is very restrictive ... it seems it doesn't have any graceful error handling ... as soon as it reads something it doesn't like, it hangs up without saying goodbye.

PR #94, opened 2019-10-24
Fixed a problem with running the elasticsearch example as updating el…

chrisdutz:
…asticsearch in general required also updating lucene. Strangely it now needed an additional external dependency.


chrisdutz:
Yeah ... no idea where that came from but without it it's looking for that class and fails.

PR #98, opened 2019-10-31
- Moved the plc4cpp, plc4net and plc4py into the sandbox

chrisdutz:
- Removed the "with-java" profile
- Adjusted the documentation
- Added documentation to the latest mspec changes


JulianFeinauer:
@chrisdutz just checked everything and I like it. So I would merge it, if nobody else vetoes?

PR #99, opened 2019-11-01
add iotdb integration example. for just saving one field for a PLC.

jixuan1989:
Hi, I finished a simple example to using plc4j and edgent to collect data and then storing the data into iotdb.

It is a simple example, but just for showing how to use IoTDB and Plc4j together.

Latter, I think I can do some integration with plc4j-opm with IoTDB. e.g., add a method in OPM entity called `createIoTDBSchema()` to create the schema in IoTDB according to the attributes of the entity. Then an example about using OPM and iotdb will be easy to be implemented.

### DIFF
--- a/plc4j/examples/hello-integration-iotdb/pom.xml
+++ b/plc4j/examples/hello-integration-iotdb/pom.xml
@@ -1,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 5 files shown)


JulianFeinauer:
Hey @jixuan1989 the example looks good. What do you think about providing a Test to see if everything runs? I could also help you with that, if you like?

JulianFeinauer:
Oh, I just saw that you write against an existing instance.. so thats rather impossible, I guess... so it looks good to me otherwise and I will merge it. It would probably be cool to add an example to write into a tsfile, what do you think @jixuan1989 ?

JulianFeinauer:
And.. congratulations to your first plc4x PR 🎉 !

jixuan1989:
> Oh, I just saw that you write against an existing instance.. so thats rather impossible, I guess... so it looks good to me otherwise and I will merge it. It would probably be cool to add an example to write into a tsfile, what do you think @jixuan1989 ?

I will :D

PR #103, opened 2019-12-04
Fixes in ADS protocol

amrod-:
Fixed the following issue:

- The Payload2TcpProtocol did not retain the incomming ByteBuf in case a message entered fragmented and stopped parsing after one payload was parsed.
- The length check missed the header size.
- ByteBuf created in Ads2PayloadProtocol were not released.

### DIFF
--- a/plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
+++ b/plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import java.util.List;
 
 public class Payload2TcpProtocol extends MessageToMessageCodec<ByteBuf, ByteBuf> {
+    private ByteBuf retainingBuf = Unpooled.EMPTY_BUFFER;
 
     private static final Logger LOGGER = LoggerFactory.getLogger(Payload2TcpProtocol.class);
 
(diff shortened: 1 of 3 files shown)


chrisdutz:
I think Julian as our Netty guru should review

plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
```diff
@@ -49,21 +50,24 @@ protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteB
...
+            TcpLength packetLength = TcpLength.of(retainingBuf);
+            AmsTcpHeader amsTcpHeader = AmsTcpHeader.of(packetLength);
+            LOGGER.debug("AMS TCP Header {}", amsTcpHeader);
+
+            out.add(retainingBuf.readBytes((int) packetLength.getAsLong()));
         }
-        if (byteBuf.readableBytes() < byteBuf.getUnsignedIntLE(AmsTcpHeader.Reserved.NUM_BYTES)) {
-            // wait till we have a complete ADS packet
-            return;
+        if (retainingBuf.readableBytes() == 0) {
```
JulianFeinauer: Could this lead to a memory leak? We should also release the byteBuf, right? Currently retain is called (L54) but never release.

plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
```diff
@@ -49,21 +50,24 @@ protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteB
             return;
         }
         LOGGER.trace("(-->IN): {}, {}, {}", channelHandlerContext, byteBuf, out);
-        if (byteBuf.readableBytes() < AmsTcpHeader.Reserved.NUM_BYTES + TcpLength.NUM_BYTES) {
-            // wait till we can read the length header
-            return;
+
+        retainingBuf = Unpooled.wrappedBuffer(retainingBuf, byteBuf.retain());
```
JulianFeinauer: Why do we need this other Buf? Why not just use the byteBuf directly?
and if its in the case of not being able to read, then netty provides the possibility to "mark" a position and "rollback" to it, see here https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#markReaderIndex--

JulianFeinauer:
@chrisdutz thanks for the hint, I added one comment :)

plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
amrod-: Every time decode is called it contains a new buffer, all buffers need to be collected until there is sufficient data to parse a message. The buffer needs to be retained the documentation specifies that after the function returns the buffer is released see https://netty.io/4.0/api/io/netty/handler/codec/MessageToMessageDecoder.html . By wrapping the buffers it is possible to keep using it as a buffer. Another option would be to copy the data to a buffer but that sounds less efficient.

plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
amrod-: The documentation: https://netty.io/4.0/api/io/netty/buffer/Unpooled.html#wrappedBuffer-io.netty.buffer.ByteBuf...-
States that the reference count ownership is transfered to this method. So when releasing (L68) the byteBuf should be released.

plc4j/protocols/ads/src/main/java/org/apache/plc4x/java/ads/protocol/Payload2TcpProtocol.java
JulianFeinauer: But the Buffer fills up until its consumed, right? Probably I dont get it right.

amrod-: One can fill a Buffer by writing to it. However in this case you get a new buffer every time, at least I haven't found anything in the documentation about that buffer being filled if not used. It would also contradict the documentation about being able to pass through the buffer if an underlying layer would later modify that same buffer.

JulianFeinauer:
Sorry @amrod- I was busy. I will go trhough it tomorrow in deep and then we can hoepfully finish and merge your work : )

sruehl:
some more potential release points:
in Ads2PayloadProtocol at the end of decode `byteBuf.release`
in Payload2SerialProtocol at the end of encode `amsPacket.release`
in Payload2TcpProtocol at the end of decode `byteBuf.release`

amrod-:
> some more potential release points:
> in Ads2PayloadProtocol at the end of decode `byteBuf.release`
> in Payload2SerialProtocol at the end of encode `amsPacket.release`
> in Payload2TcpProtocol at the end of decode `byteBuf.release`

No the super class (MessageToMessageCodec) will already call release see https://netty.io/4.0/api/io/netty/handler/codec/MessageToMessageDecoder.html

PR #105, opened 2019-12-08
Plc4 x 157 opc ua disablediscovery

nalim2:
Added the feature to skip the OPC UA discovered IP address and replace it with the provided one by the request.

+ Fix for a connection string error in the regex

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/OpcuaPlcDriver.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/OpcuaPlcDriver.java
@@ -43,7 +43,7 @@ public class OpcuaPlcDriver implements PlcDriver {
 
     public static final Pattern INET_ADDRESS_PATTERN = Pattern.compile("tcp://(?<host>[\\w.-]+)(:(?<port>\\d*))?");
     public static final Pattern OPCUA_URI_PARAM_PATTERN = Pattern.compile("(?<param>[(\\?|\\&)([^=]+)\\=([^&]+)]+)?"); //later used for regex filtering of the params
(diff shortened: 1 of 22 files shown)


plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/OpcuaPlcDriver.java
```diff
@@ -36,16 +35,15 @@ Licensed to the Apache Software Foundation (ASF) under one
  * Implementation of the OPC UA protocol, based on:
  * - Eclipse Milo (https://github.com/eclipse/milo)
  *
- * @author Matthias Milan Stlrljic
- * Created by Matthias Milan Stlrljic on 10.05.2019
+ * @author Matthias Milan Strljic
```
JulianFeinauer: Its recommended to not use @author tag on Apache Licensed files for multiple reasons, I suggest to remove it thus.

plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/ManualPLC4XOpcua.java
```diff
@@ -33,21 +33,23 @@ Licensed to the Apache Software Foundation (ASF) under one
 import java.math.BigInteger;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
+import java.util.Calendar;
```
JulianFeinauer: Why do we use last century java Date Time classes here? 👴

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/OpcuaPlcDriver.java
nalim2: Should be fixed now my lord

PR #117, opened 2020-02-07
PLC4X-174 Adjust restrictive regular expression.

splatch:


JulianFeinauer:
Hey Lukasz, this should go to rel/0.6 branch, I guess, or?

splatch:
I spotted this bug with bacnet ip driver on develop but sure, it can be backported to 0.6 branch too.

JulianFeinauer:
@chrisdutz could you also have a look here. I thought we moved all this regex stuff to one central place (using java bultin classes). This approach would not allow us to parse path parameters, right?

JulianFeinauer:
@splatch LGTM but I'm just a bit confused that the regex is there in the first place, so let me shortly check why that is than we can merge it : )
For rel/0.6 I can merge it asap.

chrisdutz:
Well I split up the parts of the connection string ... the ip/hostname stuff should now be in the tcp and udp transport modules. Just can't immediately find it but in general the url is split up into {protocol}:{transport}://{transport-config}?{parameter-string}

chrisdutz:
Aaahh ... yeah ... just had a look ... of course ... The patch is for the Transport ... so this is exactly the place where it has to be fixed.

splatch:
Since I've got thumbs up, I've pushed changes to develop. I will pray for jenkins to pass the build. ;-)

PR #118, opened 2020-02-11
Introduced a new optional "slaveId" parameter to the modbus driver

acs:


acs:
This code was not merged in develop branch and it is needed. With this PR we do it! Thanks @chrisdutz for the code and my fault not reporting after two days the results for the testing (I did it in the PLC4x ticket: https://issues.apache.org/jira/projects/PLC4X/issues/PLC4X-156) blocking the merging of this code.

PR #119, opened 2020-02-11
Update S7PlcConnection.java

generalkelly:
From my experience with the connection to a S7 the calledTsapId must contain rack and slot while the callingTsapId must not. 

This PR is refering to https://issues.apache.org/jira/projects/PLC4X/issues/PLC4X-134?filter=allopenissues


chrisdutz:
We have completely rewritten the S7 driver in the develop branch ... I think starting with the 0.7.0 releases this will no longer be an issue.

chrisdutz:
I'm going to close this one as the problem is no longer applicable to the new S7 driver.

PR #121, opened 2020-02-18
PLC4X-176 resolve issues round running on linux platforms

ottobackwards:
This pr resolves issues that were keeping the tools scripts from running on any other platform than macOS.

The fixes involved resolving warnings set by shellcheck, changing to be `bash` scripts, and using bash expression matching.

These fixes have been verified in docker containers as such: 
```bash
>  docker run -it -v `(pwd)`:/root/code ubuntu:bionic bash
>  docker run -it -v `(pwd)`:/root/code ubuntu:disco bash
>  docker run -it -v `(pwd)`:/root/code centos:7 bash
>  docker run -it -v `(pwd)`:/root/code centos:8 bash
```
and locally on macOS 10.15.3 with 
bash -> 3.2.57
bash -> 5.0.16
With 0.6.0 2

Note, that these containers do not have all the packages required to run the scripts to completion, nor the local keys, but when those packages are present the script runs without error.  If the keys are present it runs to completion.

The packages that may be missing are curl, perl-Digest-SHA.

---
This change is [Reviewable](https://reviewable.io/reviews/apache/plc4x/121)

### DIFF
--- a/tools/check_sigs.sh
+++ b/tools/check_sigs.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 ################################################################################
 ##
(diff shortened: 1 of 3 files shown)


JulianFeinauer:
👍 LGTM, thank you @ottobackwards!

PR #123, opened 2020-02-23
Use existing base image in Dockerfile

nuclearpinguin:
```
➜ docker build -t plc4x .
Sending build context to Docker daemon  20.12MB
Step 1/35 : FROM azul/zulu-openjdk:8.42.0.21 as build
manifest for azul/zulu-openjdk:8.42.0.21 not found: manifest unknown: manifest unknown
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@
 # This is the image we'll use to execute the build (and give it the name 'build').
 # (This image is based on Ubuntu)
 # Fixed version of this in order to have a fixed JDK version
-FROM azul/zulu-openjdk:8.42.0.21 as build
+FROM azul/zulu-openjdk:8 as build
 
 # Install some stuff we need to run the build
 RUN apt update -y


nuclearpinguin:
Also I've opened this [JIRA issue](https://issues.apache.org/jira/browse/PLC4X-178) because I'm facing some problems with building Docker :<

chrisdutz:
Hi,

so please help me understand the problem here. The docker image specified in the dockerfile is no longer available and by using any java 8 version the build works again?

Chris

nuclearpinguin:
> so please help me understand the problem here. The docker image specified in the dockerfile is no longer available

@chrisdutz that's correct. I could not found the tag here: https://hub.docker.com/r/azul/zulu-openjdk/tags

>by using any java 8 version the build works again?
> 

At least it starts... but I wasn't able to make it work yet.

chrisdutz:
Ah ok ... well I guess that they only host the most recent versions and remove old ones (Not that familiar with the Docker ecosystem). So I guess by just providing "8" as version it will take the most recent Java 8 version, correct?

nuclearpinguin:
> So I guess by just providing "8" as version it will take the most recent Java 8 version, correct?
 Yes

chrisdutz:
LGTM ... thanks for finding and reporting this.

PR #124, opened 2020-02-23
[WIP] Refactor PlcDriverManager as context manager

nuclearpinguin:
I would like to suggest to use `PlcDriverManager` as a context manager. In this way users don't have to care about try / excepts. I also extracted some server logic from constructor to separate method.

I will add some tests but first I have to sort out setup :)

### DIFF
--- a/sandbox/plc4py/src/main/python/org/apache/plc4x/PlcDriverManager.py
+++ b/sandbox/plc4py/src/main/python/org/apache/plc4x/PlcDriverManager.py
@@ -18,6 +18,7 @@
 import subprocess
 import time
 import warnings
+from contextlib import contextmanager
 
 from generated.org.apache.plc4x.interop.InteropServer import Client, PlcException
 from thrift.protocol import TBinaryProtocol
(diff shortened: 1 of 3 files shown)


nuclearpinguin:
@JulianFeinauer is there any particular reason for the structure of `plc4py`? It's not very python-like :)

JulianFeinauer:
@nuclearpinguin first off, hi Thomascz, nice to have you here : ) And in fact, when we started with python (no one had a clue about python), @chrisdutz and I made the structure as we know it from java / maven. So thats rather historic reason.

chrisdutz:
Hi @nuclearpinguin,

I always thought that it would be great if someone who actually knew a language would at least help with the API in order for it to feel right. So I'm fine with any changes you propose. However it would be great if it would stay possible to build the project with maven as the code generation will have to be done with maven and as long as all is located in one repo, that we can do a CI build with all parts in one run.

Chris

nuclearpinguin:
@chrisdutz can I find somewhere detailed information on how the code generation works? 

@JulianFeinauer thanks for the info, I will refactor the PLC4py. I will also familiarize myself with maven and the whole project structure.

chrisdutz:
I'd suggest starting here: https://plc4x.apache.org/developers/code-gen/index.html
And then digging though some existing drivers. 
@JulianFeinauer started writing up a tutorial, which hasn't progressed very far here:
https://plc4x.apache.org/developers/tutorials/writing-driver.html

sandbox/plc4py/src/main/python/org/apache/plc4x/PlcDriverManager.py
```diff
@@ -18,6 +18,7 @@
 import subprocess
 import time
 import warnings
+from contextlib import contextmanager
 
 from generated.org.apache.plc4x.interop.InteropServer import Client, PlcException
```
nuclearpinguin: @JulianFeinauer can you explain the import? I can't find the `generated.org...` package nowhere in the project and I am wondering what I've missed... I was able to build the project but some `target` directories appeared.

I had this info during the build:
```
-- Building without tests
-- Found FLEX: /usr/bin/flex (found version "2.6.4")
-- Found BISON: /usr/bin/bison (found version "3.0.4")
-- ----------------------------------------------------------
-- Thrift version:                               0.13.0 (0.13.0)
-- Thrift package version:                       0.13.0
--
-- Build configuration summary
--   Build compiler:                             ON
--   Build libraries:                            ON
--   Build tests:                                OFF
--   Build type:                                 RelWithDebInfo
--
-- Language libraries:
--
--   Build as3 library:                          OFF
--    - Adobe Flex compc was not found - did you set env var FLEX_HOME?
--
--   Build C++ library:                          OFF
--
--   Build C (GLib) library:                     OFF
--    - GLib missing
--
--   Build Java library:                         OFF
--
--   Build Python library:                       OFF
--    - Python libraries missing
--
--   Build Haskell library:                      OFF
--    - GHC missing
--    - Cabal missing
--
-- ----------------------------------------------------------
-- Configuring done
-- Generating done
```

But I'm not sure if this is somehow related (especially the Haskell part :D ).

chrisdutz:
The missing code is generated by thrift as part of the maven build.

PR #130, opened 2020-03-03
PLC4x-142 OPC-UA read variables

nalim2:
[fix] Changed restircitive RegEx for OPC UA node id


nalim2:
I kind of like it to get feedback ;)

For my defense there is already an OpcuaFieldTest. But those tests are missing!
i will add them and come back

PR #135, opened 2020-03-23
S7 String

etiennerobinet:
Started from dev branch and reimplemented String parsing for S7


etiennerobinet:
Sry I did an error. I am trying to start from current dev and add the S7 String parsing.

PR #139, opened 2020-03-31
Change repo URLs from incubator-plc4x to plc4x

NiklasMerz:
Gitbox does not redirect and you get bad links


chrisdutz:
You're right ... that's a relict from graduation ...

PR #141, opened 2020-04-11
PLC4X-193 Support for little endian fields

splatch:
This is work in progress to support little endian numbers in the operations. Not sure what to do with byte sequences and reserved fields. Theoretically they should work.

### DIFF
--- a/build-utils/protocol-base-mspec/src/main/antlr4/org/apache/plc4x/plugins/codegenerator/language/mspec/MSpec.g4
+++ b/build-utils/protocol-base-mspec/src/main/antlr4/org/apache/plc4x/plugins/codegenerator/language/mspec/MSpec.g4
@@ -27,10 +27,10 @@ complexTypeDefinition
  ;
 
 complexType
- : 'type' name=idExpression (LBRACKET params=argumentList RBRACKET)? fieldDefinition*
- | 'discriminatedType' name=idExpression (LBRACKET params=argumentList RBRACKET)? fieldDefinition+
- | 'enum' type=typeReference name=idExpression (LBRACKET params=argumentList RBRACKET)? enumValues=enumValueDefinition+
- | 'dataIo' name=idExpression (LBRACKET params=argumentList RBRACKET)? dataIoTypeSwitch=dataIoDefinition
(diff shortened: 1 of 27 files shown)


build-utils/protocol-base-mspec/src/main/antlr4/org/apache/plc4x/plugins/codegenerator/language/mspec/MSpec.g4
```diff
@@ -152,6 +152,10 @@ dataType
  | base='dateTime'
  ;
 
+ENDIANNESS_LITERAL
+ : 'big endian' | 'little endian'
+ ;
+
```
chrisdutz, line 158: Could we make this "LE" or "BE"?

build-utils/protocol-base-mspec/src/main/java/org/apache/plc4x/plugins/codegenerator/language/mspec/model/references/DefaultIntegerTypeReference.java
```diff
@@ -23,8 +23,16 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+
+    public DefaultIntegerTypeReference(SimpleBaseType baseType, int sizeInBits, boolean littleEndian) {
         super(baseType, sizeInBits);
+        this.littleEndian = littleEndian;
+    }
+
+    @Override
+    public boolean isLittleEndian() {
+        return littleEndian;
     }
```
chrisdutz, line 36: We could consider removing the "@Override" and leaving the build-tools repo unchanged ...

build-utils/protocol-base-mspec/src/main/antlr4/org/apache/plc4x/plugins/codegenerator/language/mspec/MSpec.g4
splatch: Not sure if you want LE/BE instead of little endian and big endian, or next to i? I took longer notation as it is self descriptive. Making BE/LE option next to these isn't a problem as I extracted comparison to one place which works with antlr token regardless of its use context (type, field etc).

splatch:
I'm closing this one since it is not required to get ADS handling. Basic fields which are assembled to bit, byte or int types can be mapped using a receiving order without need to swap byte positions.

PR #153, opened 2020-05-11
Feature/c api subscriptions

chrisdutz:
Working branch for subscription support ...

### DIFF
--- a/sandbox/plc4c/api/include/plc4c/subscribe.h
+++ b/sandbox/plc4c/api/include/plc4c/subscribe.h
@@ -1,0 +1,189 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef PLC4C_SUBSCRIBE_H_
+#define PLC4C_SUBSCRIBE_H_
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "plc4c/types.h"
+
+/*
+ *
+ * Subscriptions
+ *
+ */
+
+/**
+ * Destroys a given subscription-request.
+ *
+ * @param subscription_request the subscription-request.
+ */
+void plc4c_subscription_request_destroy(
+    plc4c_subscription_request *subscription_request);
+
+/**
+ * Returns the plc4c_connection for a give subscription request
+ * @param subscription_request plc4c_subscription_request
+ * @return plc4c_connection
+ */
+plc4c_connection *plc4c_subscription_request_get_connection(
+    plc4c_subscription_request *subscription_request);
+
+/**
+ * Sets the plc4c_connection for a given subscription request
+ * @param subscription_request plc4c_subscription_request
+ * @param connection plc4c_connection
+ */
+void plc4c_subscription_request_set_connection(
+    plc4c_subscription_request *subscription_request,
+    plc4c_connection *connection);
+
+/**
+ * Actually executes the subscription-request.
+ * @param connection connection this subscription-request will be executed on.
+ * @param subscription_request the subscription-request object.
+ * @param subscription_request_execution pointer to a data-structure handling
+ * one execution of the subscription-request.
+ * @return plc4c_return_code
+ */
+plc4c_return_code plc4c_subscription_request_execute(
+    plc4c_subscription_request *subscription_request,
+    plc4c_subscription_request_execution **subscription_request_execution);
+
+/**
+ * Destroys a given subscription-request execution.
+ *
+ * @param subscription_request_execution the subscription-request execution.
+ */
+void plc4c_subscription_request_execution_destroy(
+    plc4c_subscription_request_execution *subscription_request_execution);
+
+/**
+ * Check if the subscription-request is completed successfully.
+ *
+ * @param subscription_request_execution the subscription-request execution.
+ * @return true if the subscription-request is completed successfully.
+ */
+bool plc4c_subscription_request_execution_check_finished_successfully(
+    plc4c_subscription_request_execution *subscription_request_execution);
+
+/**
+ * Check if the subscription-request is completed unsuccessfully.
+ *
+ * @param subscription_request_execution the subscription-request execution.
+ * @return true if the subscription-request is completed with an error.
+ */
+bool plc4c_subscription_request_execution_check_finished_with_error(
+    plc4c_subscription_request_execution *subscription_request_execution);
+
+/**
+ * Retrieve the subscription-response from a given subscription-request
+ * execution.
+ *
+ * @param subscription_request_execution the subscription-request execution.
+ * @return the subscription-response.
+ */
+plc4c_subscription_response *plc4c_subscription_request_execution_get_response(
+    plc4c_subscription_request_execution *subscription_request_execution);
+
+/*
+ *
+ * Unsubscriptions
+ *
+ */
+
+/**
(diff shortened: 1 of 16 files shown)


sandbox/plc4c/api/include/plc4c/subscribe.h
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Sets the plc4c_connection for a given subscription request
+ * @param subscription_request plc4c_subscription_request
+ * @param connection plc4c_connection
+ */
+void plc4c_subscription_request_set_connection(
+    plc4c_subscription_request *subscription_request,
+    plc4c_connection *connection);
+
```
ottobackwards, line 57: I don't understand the difference between these three things.  Can you explain them?

sandbox/plc4c/api/include/plc4c/subscribe.h
```diff
@@ -0,0 +1,244 @@
...
+bool plc4c_subscription_check_data_available(void *subscription_handle);
+
+/**
+ * Destroys a given subscription-request execution.
+ *
+ * @param subscription_request_execution the subscription-request execution.
+ */
+void plc4c_subscription_request_execution_destroy(
+    plc4c_subscription_request_execution *subscription_request_execution);
+
```
ottobackwards, line 116: I am not sure you would use two methods here, why not one that returns a boolean or an enum?  When would you call one and not the other?

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+        if (plc4c_connection_get_connected(connection)) {
+          printf("SUCCESS\n");
+          state = CONNECTED;
+        } else if (plc4c_connection_has_error(connection)) {
+          printf("FAILED\n");
+          return -1;
+        }
+        break;
+      }
+      case CONNECTED: {
```
ottobackwards, line 124: I think you want to update the comments and explain the subscription stuff you are doing

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+            subscription_request, &subscription_request_execution);
+        if (result != OK) {
+          printf("FAILED\n");
+          return -1;
+        } else {
+          state = SUBSCRIPTION_REQUEST_SENT;
+        }
+        break;
+      }
+        // Wait until the read-request execution is finished.
```
ottobackwards, line 157: This makes it seem like the request being sent is equal to the response being received
I think the state would be SUBSCRIPTION_WAITING_RESPONSE

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+            plc4c_subscription_request_execution_check_finished_with_error(
+                subscription_request_execution)) {
+          printf("FAILED\n");
+          return -1;
+        }
+        break;
+      }
+      case SUBSCRIPTION_RESPONSE_RECEIVED: {
+        // Get the response for the given subscription-request.
+        plc4c_subscription_response *subscription_response =
```
ottobackwards, line 173: This would be _check_resonse

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+          return -1;
+        }
+
+        // Iterate over all returned items.
+        plc4c_list_element *cur_element =
+            plc4c_utils_list_head(subscription_response->response_items);
+        while (cur_element != NULL) {
+          plc4c_response_subscription_item *subscription_item_item =
+              cur_element->value;
+
```
ottobackwards, line 187: I think there would be value, in letting the caller set a 'cookie' or identifier in the subscription that they could check when they got it back.
It may not be a char* string that they have to strcmp.  This is common.

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+          if (subscription_item_item->response_code == OK) {
+            // Save the subscription handles...
+            subscription_handle = subscription_item_item->subscription_handle;
+          }
+
+          cur_element = cur_element->next;
+        }
+
+        // Clean up.
```
ottobackwards, line 200: So this just reads one and then quits?  Wouldn't we want them to read until someone hit a key?

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
```diff
@@ -0,0 +1,264 @@
...
+        plc4c_subscription_response_destroy(subscription_response);
+        plc4c_subscription_request_execution_destroy(
+            subscription_request_execution);
+        plc4c_subscription_request_destroy(subscription_request);
+
```
ottobackwards, line 205: I don't am not sure about separating the states here.  Just as I don't understand all the separation of concerns with the item types.
Even if you add multiple types of items, in the end, you would just check for data.  I am not sure why you would check for poll, check for events.  Just check for data.
If they got a cookie set, they can tell which it was....

But I may be missing something, can you explain a bit?

sandbox/plc4c/spi/include/plc4c/spi/types_private.h
```diff
@@ -178,7 +198,20 @@ struct plc4c_request_value_item_t {
 struct plc4c_response_value_item_t {
   plc4c_item *item;
   plc4c_response_code response_code;
-  void *value;
+  plc4c_data *value;
+};
+
```
ottobackwards, line 203: This maybe where we want to have the 'cookie'

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
chrisdutz: No ... in this case we only have one subscription item, so it saves the subscription handle and then goes to the next step and waits till it receives 10 events and then exits ...

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
chrisdutz: The point of this example is not to build a good application, but to demonstrate the API ... so this counter demonstrates the api usage without any constraints about how to handle key presses.

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
chrisdutz: Well the protocols usually support these and they are handled completely different. One is some times polling some times a cyclic push ... some protocols support updates on value changes ... in this the PLC actively sends data if something changes. The third kind are events that are not related to values but to events happening in the PLC program. So they are completely different forms of resources or update patterns.

sandbox/plc4c/spi/include/plc4c/spi/types_private.h
chrisdutz: Don't understand this cookie thing ...

sandbox/plc4c/api/include/plc4c/subscribe.h
ottobackwards: Cyclic item, Event Item, Change of state item

sandbox/plc4c/api/include/plc4c/subscribe.h
ottobackwards: I mean that it is confusing how check_finished_successfully  == false is not the same as check_finished_with_error == true

sandbox/plc4c/examples/hello-subsciptions/src/hello_subscriptions.c
ottobackwards: Ok, I can see then why you would care about that when you configured the item.  But why when checking for data?  The caller doesn't need to know do they?

So add_xxx_item()
add_yyy_item()
subscribe()
get_data()
get_data()
get_date().....

when you get the data, you can check the type or which item returned

sandbox/plc4c/spi/include/plc4c/spi/types_private.h
ottobackwards: I can't find an example, so let's forget it.
It is basically that you would want to store come context item or information in the request item, so that when you got the callback you could identify the object and use the context instead of having to map the request item to something else elsewhere

ottobackwards:
I'm fine to merge this and move on

PR #155, opened 2020-05-12
PLC4X-197 do not stop the global timer, remove and cancel the Timeouts instead

ottobackwards:
rat exception for cmake-* directories' as well.

SingleItemToSingleRequestProtocol uses the global DefaultNettyPlcConnection timer object, which is a static singleton shared between all instances.

It was calling `stop()` on the timer, which cancels all the outstanding tasks in the timer when the channel when inactive.  This has the effect of killing all the other connections.

This pr changes SingleItemToSingleRequestProtocol to instead, cancel all it's outstanding Timeouts and not touch the timer.

---
This change is [Reviewable](https://reviewable.io/reviews/apache/plc4x/155)


ottobackwards:
I am open to suggestions on how to write a test for this

chrisdutz:
LGTM

PR #156, opened 2020-05-14
changes to allow running tests not coupled to maven build phase

ottobackwards:
If you are just working on c, and in CLion or something, the dependency on BUILD_PHASE for the maven caller doesn't make sense.

Changes so that you can work only in C.

note: you still have to run mvn once to get unity to install in target.

---
This change is [Reviewable](https://reviewable.io/reviews/apache/plc4x/156)


chrisdutz:
The thing with this is that now you can't build without tests ... I intentionally built it that way, that you can do a maven build and skip tests ...

chrisdutz:
Perhaps inverting the logic and allowing to disable tests would be an option ...

ottobackwards:
OK, I'll add that.

ottobackwards:
never mind

PR #158, opened 2020-05-25
Updated Eclipse Milo to 0.3.7

patrickse:
Eclipse Milo 0.3.7 supports relaxed nonce validation when not using protected credentials. It's needed to talk to development PLC from Schneider Electric.

https://github.com/eclipse/milo/commit/35ee06ea87676db9fdf4caf748b9265c40d1c841

Signed-off-by: Patrick Sernetz <psernetz@cad-schroer.de>

### DIFF
--- a/plc4j/drivers/opcua/pom.xml
+++ b/plc4j/drivers/opcua/pom.xml
@@ -94,17 +94,17 @@
     <dependency>
       <groupId>org.eclipse.milo</groupId>
       <artifactId>sdk-client</artifactId>
-      <version>0.3.6</version>
+      <version>0.3.7</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.milo</groupId>
       <artifactId>stack-core</artifactId>
-      <version>0.3.6</version>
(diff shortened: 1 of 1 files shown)


JulianFeinauer:
Thanks @patrickse ! Do we have integration tests for milo or did you test the code already?

patrickse:
I am not aware of any integration test. We've changed the dependency and deployd it to our platform. It looks ok for now. How could I implement an integration test?

JulianFeinauer:
Good question, indeed. But good to know that it works for you. @nalim2 we should definetly have a look there. Probably we could start an OPC Server and start a Client and just do some sending. Probably even using the awesome TestContainers Project?

nalim2:
Yeah i think to include there a Sample Server should be a good idea. I would first just integrate a Milo example server and would transfere the ManualTest class to an integration test or what you mean by the TestContainers Project?

If there would be an alternativ server it would be better to get possible specification problems

nalim2:
But just the updated versions lgtm

PR #161, opened 2020-05-27
C-Code-Generation (Please don't merge ... this is used for allowing code-reviews)

chrisdutz:
…) so we can do code reviews on it.


ottobackwards (commented):
Looks good,  I think we should have the camelToSnake utility look ahead and group multiple cap letter together

chrisdutz:
Yeah that would indeed shorten some names by 50%

chrisdutz:
So how about if the next letter is uppercase too, treat the current as the same block.

ottobackwards:
Yes something like that

ottobackwards:
That looks great!

sandbox/plc4c/spi/include/plc4c/spi/read_buffer.h
```diff
@@ -0,0 +1,80 @@
...
+
+struct plc4c_spi_read_buffer {
+
+};
+typedef struct plc4c_spi_read_buffer plc4c_spi_read_buffer;
+
+uint32_t plc4c_spi_read_get_pos(plc4c_spi_read_buffer* buf);
+
+uint32_t plc4c_spi_read_get_total_bytes(plc4c_spi_read_buffer* buf);
+
```
ottobackwards, line 34: bytes not bits?

chrisdutz: Well these are not methods used for the normal parsing. All of the normal read methods are bit-based

ottobackwards:
wrt read/write buffer.  I think we'll need something for endianness and float formats

chrisdutz:
;-) ... I am currently replicating what we have in Java and here the endianess is handled internally. It's handled in the constructor of the buffer.

sandbox/plc4c/spi/include/plc4c/spi/read_buffer.h
ottobackwards: Can you point me to an example reading bits?
Why would would you pass in num_bits to read a field of a certain size like float?  Are you saying that not only is the buffer abstracting reading types from a byte array, but is also abstracting reading types from a byte array with some packing scheme?

Maybe you should put something in the headers about what you are going after?

I think there should be methods for reading just a byte* of n bytes out as well

chrisdutz: in this case you would have an "array" typed field of type "byte" ...

ottobackwards: I'll wait to look at this until it is commented, so I can understand what you are doing.

PR #166, opened 2020-06-16
Team addition strljic

nalim2:


JulianFeinauer (approved):
To be honest I would prefer a picture of you without a tie but... better than nothing, I guess : )

nalim2:
Haha ok i will update that one later for you <3

PR #168, opened 2020-06-16
Feature/c code generation tagged unions

chrisdutz:
This is the new approach using tagged unions instead of separate types for every sub-type.


build-utils/language-c/src/main/java/org/apache/plc4x/language/c/CLanguageTemplateHelper.java
```diff
@@ -436,8 +476,45 @@ private String toExpression(Field field, Term term, Function<Term, String> varia
         }
     }
 
-    public String toVariableParseExpression(Field field, Term term, Argument[] parserArguments) {
+    public String toVariableParseExpression(ComplexTypeDefinition baseType, Field field, Term term, Argument[] parserArguments) {
         VariableLiteral vl = (VariableLiteral) term;
```
ottobackwards, line 549: magic string

sandbox/plc4c/generated-sources/modbus/src/modbus_constants.c
```diff
@@ -24,18 +24,18 @@
 #include "modbus_constants.h"
 
 // Parse function.
-plc4c_return_code plc4c_modbus_read_write_modbus_constants_parse(plc4c_spi_read_buffer* buf, plc4c_modbus_read_write_modbus_constants** message) {
+plc4c_return_code plc4c_modbus_read_write_modbus_constants_parse(plc4c_spi_read_buffer* buf, plc4c_modbus_read_write_modbus_constants** _message) {
```
ottobackwards, line 27: to me, when I see _var_name I think of a private var, not only in python but in some c-ish languages I've worked in.  is there some standard that you are writing two using _ in a parameter name?

ottobackwards:
that is too much to look through all of it.
It looks good.  A couple of things that come to mind:

- can we run clang-format after generation?
- should we be generating some logging?
- _message as a parameter is strange to _me_
- so many (*_message)-> everywhere, can we use a typedef or something?

None of these things are for this PR

build-utils/language-c/src/main/java/org/apache/plc4x/language/c/CLanguageTemplateHelper.java
chrisdutz: ??

sandbox/plc4c/generated-sources/modbus/src/modbus_constants.c
chrisdutz: Well I had to somehow make sure the names a user can choose in mspec don't interfere with those used internally ... so I prefixed them with a "_" if you have another suggestion, I'm  fine with changing.

chrisdutz:
> that is too much to look through all of it.
> It looks good. A couple of things that come to mind:
> 
> * can we run clang-format after generation?
> * should we be generating some logging?
> * _message as a parameter is strange to _me_
> * so many (*_message)-> everywhere, can we use a typedef or something?
> 
> None of these things are for this PR

Well I would rather have the code generated in a way that it doesn't require formatting ...

I didn't generate any logging as I didn't know of any logging framework to use ... we could whip up a skeleton of functions that we can use for logging and plug in a "printf" logger or whatever we like to do the actual logging ... that might be a good idea.

As I mentioned before ... I wanted to ensure the internal variable names don't clash with the ones a user can use in mspec ... people tent to get anoyed if they have to use other named due to internal stuff.

Why do we need to typedef that ... It's not that anyone is ever going to write that code ... but if you like, of course that's probably a really simple change.

ottobackwards:
@chrisdutz that is fair enough.  no reason not to keep it like it is

build-utils/language-c/src/main/java/org/apache/plc4x/language/c/CLanguageTemplateHelper.java
ottobackwards: "CAST" could be a constant.  If there are other things you could do that too.

chrisdutz: Yeah ... correct ... but I think there's only one place where it's used ...

PR #170, opened 2020-07-08
PLC4X-207 When a Handler Timeout occurs cancel the read future to not…

JulianFeinauer:
…ify the user.


ottobackwards:
This looks good.  It would be nice if there were some tests for this though

+1

JulianFeinauer:
@ottobackwards I agree with you (I also would like to have a test), yet I'm unsure how we can provide a Test with lower complexity then the implementation. Perhaps @chrisdutz has an idea with his setup and the test suite?

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
```diff
@@ -299,6 +319,14 @@ public void onConnect(ConversationContext<TPKTPacket> context) {
         return future;
     }
 
+    /**
+     * This method is only called when there is no Response Handler.
+     */
+    @Override
+    protected void decode(ConversationContext<TPKTPacket> context, TPKTPacket msg) throws Exception {
+        throw new IllegalStateException("This should not happen!");
```
chrisdutz, line 300: Well it actually could happen ... yesterday I noticed that error responses would go here ... but it's a good place for starting to implement some sort of error handling.

chrisdutz (approved):
LGTM

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
JulianFeinauer: @chrisdutz yes. BUT the issue is that we want to notify people waiting on the response which is another thread which waits for a future.
One slution would be to have a set of all open futures for this connection and cancel all of them if this method is called ("undefined behavior"). I took the other way and notify the request / reply pattern when a connection si dropped (but nonetheless the paket will then  go there).

chrisdutz: I wasn't objecting ... I hope it didn't appear that way ... was more a pointer for myself to continue here to implement correct handling of errors.

chrisdutz:
> @ottobackwards I agree with you (I also would like to have a test), yet I'm unsure how we can provide a Test with lower complexity then the implementation. Perhaps @chrisdutz has an idea with his setup and the test suite?

Hi ... I think you should have a look at the XML-base IT framework ... here I can add delays which would be a perfect match for implementing such a test.

Please check out:
plc4j/drivers/s7/src/test/resources/testsuite/S7DriverIT.xml
Where I would also suggest to add the test to.

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
chrisdutz: Just today did I push some changes that should help greatly with the decoding of errors in S7 ... the main problem was, that in case of an error, the response is not a S7MessageResponseData type, but a S7MessageResponse (without Data) ... I changed the code to wait for a S7Message with the given PDU Id and to handle that then ... so I think most bad errors were due to this ... perhaps the problem now will go away.

chrisdutz:
So right now I think we could probably add a test ... I do not want to force Julian to dig into my IT framework ... if you could explain the situation in which this code would kick in, I could probably come up with a testcase.

splatch:
I actually re-implemented your fixes in #203 . :-/

splatch:
I've pushed changes I made for other PR hence this one needs to be rebased/updated.

splatch:
I've updated code so it works with recent develop. Now S7 driver should properly handle timeouts.

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
```diff
@@ -69,7 +72,7 @@
  */
 public class S7ProtocolLogic extends Plc4xProtocolBase<TPKTPacket> {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(S7ProtocolLogic.class);
```
chrisdutz, line 72: as it's a static, I would prefer the LOGGER ... or make it non-static, then use the lowercase version.

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
```diff
@@ -91,35 +94,50 @@ public void setDriverContext(DriverContext driverContext) {
...
+        if (context.isPassive()) {
+            logger.info("S7 Driver running in PASSIVE mode.");
+            s7DriverContext.setPassiveMode(true);
+            // No login required, just confirm that we're connected.
+            context.fireConnected();
+            return;
+        }
+
         // Only the TCP transport supports login.
-        if(!context.isPassive()) {
```
chrisdutz, line 95: Why remove this?

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
```diff
@@ -136,32 +154,27 @@ public void onConnect(ConversationContext<TPKTPacket> context) {
...
-                });
-        }
-        // This usually when we're running a passive mode river.
-        else {
-            LOGGER.info("S7 Driver running in PASSIVE mode.");
-            s7DriverContext.setPassiveMode(true);
-
-            // No login required, just confirm that we're connected.
-            context.fireConnected();
-        }
```
chrisdutz, line 164: We should keep this as we are planning on making the drivers passive capable.

chrisdutz (changes requested):
Would be cool if you could fix the naming of the logger variable and re-add the exception for passive mode drivers

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
splatch: It is moved up in the method body so we do not have bigger block with active connection under if brackets. See lines 97-104.

plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
splatch: Good spot, I did not notice it was static. AFAIR slf4j and modern log4j (even 1.2) recommendation is to use final without static classifier.

chrisdutz (approved):

JulianFeinauer:
@splatch can this PR be closed and the branch be deletet? I think you did all that already in #203 , right?

PR #171, opened 2020-07-16
[PLC4X-216]update IoTDB JDBC example and session API example; add the related doc on website

jixuan1989:
Hi, 
this pr provides using IoTDB session API for writing data. 
The related doc on the website is also supplied.


chrisdutz:
Probably might be a good idea to add the "useJDBC" to the options and allow switching from the command-line?

jixuan1989:
> Probably might be a good idea to add the "useJDBC" to the options and allow switching from the command-line?

good idea. done.

PR #172, opened 2020-07-17
Feature/modbus add additional address formats and change lowest register to 1.

hutcheb:
- Address formats similar to 400001, 000001 were added
- Address formats similar to 4x00001, 0x000001 were added.
- Changed the lowest address the user can reference to 1 instead of 0.

### DIFF
--- a/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
+++ b/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
@@ -28,26 +28,23 @@ import java.util.regex.Pattern;
 public abstract class ModbusField implements PlcField {
 
     public static final Pattern ADDRESS_PATTERN = Pattern.compile("(?<address>\\d+)(\\[(?<quantity>\\d+)])?");
+    protected static final int protocolAddressOffset = 1;
 
     private final int address;
 
     private final int quantity;
 
     public static ModbusField of(String addressString) throws PlcInvalidFieldException {
-        Matcher matcher = ModbusFieldCoil.ADDRESS_PATTERN.matcher(addressString);
-        if(matcher.matches()) {
+        if(ModbusFieldCoil.matches(addressString)) {
             return ModbusFieldCoil.of(addressString);
         }
-        matcher = ModbusFieldDiscreteInput.ADDRESS_PATTERN.matcher(addressString);
-        if(matcher.matches()) {
+        if(ModbusFieldDiscreteInput.matches(addressString)) {
             return ModbusFieldDiscreteInput.of(addressString);
         }
-        matcher = ModbusFieldHoldingRegister.ADDRESS_PATTERN.matcher(addressString);
-        if(matcher.matches()) {
+        if(ModbusFieldHoldingRegister.matches(addressString)) {
             return ModbusFieldHoldingRegister.of(addressString);
         }
-        matcher = ModbusFieldInputRegister.ADDRESS_PATTERN.matcher(addressString);
-        if(matcher.matches()) {
+        if(ModbusFieldInputRegister.matches(addressString)) {
             return ModbusFieldInputRegister.of(addressString);
         }
         throw new PlcInvalidFieldException("Unable to parse address: " + addressString);
@@ -55,6 +52,9 @@ public abstract class ModbusField implements PlcField {
 
     protected ModbusField(int address, Integer quantity) {
         this.address = address;
+        if ((this.address + protocolAddressOffset) <= 0) {
+            throw new IllegalArgumentException("address must be greater then zero. Was " + (this.address + protocolAddressOffset));
+        }
         this.quantity = quantity != null ? quantity : 1;
         if (this.quantity <= 0) {
             throw new IllegalArgumentException("quantity must be greater then zero. Was " + this.quantity);
--- a/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
+++ b/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
@@ -28,7 +28,8 @@ import java.util.regex.Pattern;
 public abstract class ModbusField implements PlcField {
 
     public static final Pattern ADDRESS_PATTERN = Pattern.compile("(?<address>\\d+)(\\[(?<quantity>\\d+)])?");
-    protected static final int protocolAddressOffset = 1;
+    public static final Pattern FIXED_DIGIT_MODBUS_PATTERN = Pattern.compile("(?<address>\\d{4,5})(\\[(?<quantity>\\d+)])?");
+    protected static final int PROTOCOL_ADDRESS_OFFSET = 1;
 
     private final int address;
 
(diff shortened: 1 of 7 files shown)


plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
```diff
@@ -28,33 +28,33 @@ Licensed to the Apache Software Foundation (ASF) under one
...
-        if(matcher.matches()) {
+        if(ModbusFieldInputRegister.matches(addressString)) {
             return ModbusFieldInputRegister.of(addressString);
         }
         throw new PlcInvalidFieldException("Unable to parse address: " + addressString);
     }
 
     protected ModbusField(int address, Integer quantity) {
         this.address = address;
+        if ((this.address + protocolAddressOffset) <= 0) {
```
chrisdutz: Are you sure this shouldn't be "this.address - protocolAddressOffset" ... as far as I remember something "400001" would refer to address holding-register address 0 ...

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
```diff
@@ -28,33 +28,33 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+            throw new IllegalArgumentException("address must be greater then zero. Was " + (this.address + protocolAddressOffset));
```
chrisdutz: Same as above

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldCoil.java
```diff
@@ -26,17 +26,37 @@ Licensed to the Apache Software Foundation (ASF) under one
 public class ModbusFieldCoil extends ModbusField {
 
     public static final Pattern ADDRESS_PATTERN = Pattern.compile("coil:" + ModbusField.ADDRESS_PATTERN);
+    public static final Pattern ADDRESS_SHORTER_PATTERN = Pattern.compile("0" + ModbusField.ADDRESS_PATTERN);
```
chrisdutz: I would probably use a second ADDRESS_PATTERN here as this would also mach "01" ... I think it has to be 5 digits, correct? So perhaps using a "FIVE_DIGIT_MODBUS_PATTERN" instead would be a good idea?

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldDiscreteInput.java
```diff
@@ -26,21 +26,40 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+        } else {
+          throw new PlcInvalidFieldException(addressString, ADDRESS_PATTERN);
         }
-        int address = Integer.parseInt(matcher.group("address"));
+        return matcher;
+    }
+
+    public static ModbusFieldDiscreteInput of(String addressString) throws PlcInvalidFieldException {
+        Matcher matcher = getMatcher(addressString);
+        matcher.find();
```
chrisdutz: You are ignoring the result of the find ... what's it meant for?

chrisdutz: perhaps a "matches(...)" check at the beginning would be good ... I mean you do the check in the FieldHandler, but one could call the of method from somewhere else ... so instead of "matcher.find()" I would probably do a "if(!matcher.matches())" and throw an InvalidFieldException if it doesn't match.

chrisdutz (changes requested):
Hi Ben,

thanks for this PR. I really like it a lot. I did have some little suggestions though.
- In the FieldHandler I think the offset should be subtracted, not added
- For the new formats, I think a pattern checking for 5-digit numbers would be a good thing
- Calling "check" instead of "find" to validate in the "of" methods.

But I really like the look of your changes. Thanks a lot for that :-)

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusField.java
hutcheb: this.address stores the address sent over the wire starting at 0 instead of address 1. I didn't want to have a check that checks if the value is < 0 and have the exception say you can't have an address < 1 as it might be confusing.

In the constructor for each memory area I subtract 1 from the address before storing it in this.address.

chrisdutz: Aaaahh yeah ... now I got you ... and yes ... indeed "address" would be the decremented value ... so yeah ... just ignore my comment ;-)

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldCoil.java
hutcheb: I wasn't too sure about this because depending on the device they will list the Modbus address as 40001 or 400001. The extra 0 seems to have been added when the larger memory areas started to be used.

In Proworx NXT when referencing Modbus addresses you can just type 41 and it will always take the first digit as the memory area, the rest will be the address. I wasn't sure about extending the format to this extent.

It doesn't make sense to have more than 5 digits for the address though or values greater than 65536. I can fix this up.

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldDiscreteInput.java
hutcheb: Calling the getMatcher mehod checks for the case that the pattern isn't found by raising the exception. I moved the exception to that method so I didn't have to worry about returning a null value.

I found that the find() call then needs to be added as I wasn't calling the matches() method to find the first match. We can ignore the result of this as the getMatcher method has already confirmed that at least one instance will be found.

chrisdutz: Yeah .. in that case I would prefer "5 or more" digits (You can probably just do 5-6 digits)

chrisdutz: Ups seems that response wen't to the wrong comment ... sorry ... well I just noticed that IntelliJ was complaining ... And yeah ... if it wouldn't match, it wouldn't pass the getMatcher() call ... so how about instead of creating one matcher to check if it matches and then to return a new one ... how about returning the instance you used to check?

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldCoil.java
chrisdutz: Yeah .. in that case I would prefer "5 or more" digits (You can probably just do 5-6 digits)

plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/field/ModbusFieldDiscreteInput.java
hutcheb: Yeah that makes more sense, I'll change that.

PR #175, opened 2020-08-02
See issue reported by Stefano Bossi.

JulianFeinauer:
- removed assert as this is not called in any java setup in S7Field
- Added support for parsing of multiple requested fields.


chrisdutz:
LGTM

PR #177, opened 2020-08-10
Feature/port simotion rel 06

JulianFeinauer:
Ported Simotion Support code from @chrisdutz to 0.6 branch and replaced mspec part with "native" logik.


sruehl:
do we still need this PR? Looks pretty empty to me atm

PR #178, opened 2020-08-14
Fix ToAnsi method : bufferoverflow

JohannaMillet:
ToAnsi metod in EipProtocolLogic throws Bufferoverflow exception for tags that have more than one '.' in it.
Fixed by increased allocated data in Buffer.


ottobackwards:
Thanks for the contribution.
Is there a way you can write a unit test for this issue?

chrisdutz:
Hi Johanna, 

sorry for me picking this up so late ... I just had a look ... your code changes generally look good, and I'm not going to insist on tests as in that case I would have to reject most of my stuff to ;-)

Does the driver work with your changes? Are you able to read with multiple "." in them? I'm asking this stupid questions as I don't know the protocol and want to understand if this just makes the Execption go away.

Chris

PR #179, opened 2020-08-16
Feature/plc4py

vemmert:
I set the minimum Python version to 3.5, because that's what the current pytest requires.


ottobackwards:
I don't understand the commits from 2019?  Can you rebase this into a single commit so we can start clean?
Also, i don't think we need the egg info

ottobackwards:
Can we use python 3.8?  3.9 is about to come out.  We shouldn't start off so behind.

ottobackwards (approved):
looks good, we can iterate on the rest.
I was able to ->
- pyenv local 3.8.2
- poetry install
- poetry pytest

PR #182, opened 2020-08-28
PLC4X-244 Support for variable length padding fields.

splatch:
Padding expression must now evaluate to an integer specifying how many times 'padding' field needs to be read or written to the buffer.

S7 parser tests are fine with this change as well as my experiments with socketcan payloads.


chrisdutz:
I will probably need to add something to the size-calculation in C too, but I'll take care of this. Perhaps also change the name of the varialbe from "needsPadding" to "numPading" or similar ... "needs" sounds binary to me.

chrisdutz:
Just saw that due to the nature of C ... you found all the places that needed adjusting ... thank you very much :-)

chrisdutz (approved):

splatch:
Good spot on naming, I did a blind change trying to keep it close to initial names. Anyhow, thanks for taking care of this!

PR #191, opened 2020-09-23
Implement WriteBuffer float, double and BigInteger functions

hutcheb:
Initial commit, hasn't been tested at all.


plc4j/drivers/knxnetip/src/main/java/org/apache/plc4x/java/knxnetip/utils/KnxHelper.java
```diff
@@ -46,9 +46,16 @@ public static void f16toBytes(WriteBuffer io, Object param) {
             boolean negative = value < 0;
             final int exponent = Math.getExponent(value);
             final double mantissa = value / Math.pow(2, exponent);
+            //io.writeBit(negative);
+            //io.writeInt(4, exponent);
+            //io.writeDouble(11, Math.getExponent(mantissa), mantissa); //Don't think this works BH
+
+            String mantissaString = Double.toString(mantissa);
+            int mantissaInteger = Integer.parseInt(mantissaString.substring(mantissaString.indexOf('.')));
```
chrisdutz, line 54: What's the String conversion for? I think in this case a simple double-to-int-cast should do the trick ....

chrisdutz (approved):
LGTM :-) Thanks

plc4j/drivers/knxnetip/src/main/java/org/apache/plc4x/java/knxnetip/utils/KnxHelper.java
hutcheb: The string conversion is to get the digits after the decimal place of the mantissa as part of writing a 16-bit float.
(In my head the mantissa is always going to have a one before the decimal place which can be discarded)
The digits after are then converted to an integer to be written as the manitssa part.

I haven't tested this part of it, it has a slight chance of working for most cases but not all. I'll create a 16-bit floating point PlcValue at some point and test it with Modbus and then move it over to the KNX protocol as a custom PlcValue.

hutcheb:
Little Endian BigInteger functions havn't been tested.

PR #192, opened 2020-09-30
Refactor Field Handler Classes

hutcheb:
First attempt at refactoring the field handler class, currently troubleshooting issue with new PlcValue of method.


hutcheb:
StaticSerialzer when writing booleans needs some work.

hutcheb:
Bit of a mismatch of a bunch of issues:-
- Implemented WriteBuffer - Big Integer and floating point functions.
- Implemented staticserialzer support for PlcLists
- Moved Modbus to use the static serialiser so as not to duplicate serializing code
- Refactored the field handler code, esp for Modbus, no uses a class lookup within PlcValues. Other protocols haven't been tested.
- Added support within the hello world write example for writing arrays.
- Fix for bit lengths within the WriteBuffer writeXXX functions.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
```diff
@@ -169,6 +169,30 @@ public PlcBYTE(@JsonProperty("value") short value) {
...
+    public boolean isByte() {
+        return (value != null) && (value <= Byte.MAX_VALUE) && (value >= Byte.MIN_VALUE);
+    }
+
+    @Override
+    @JsonIgnore
+    public byte getByte() {
+        return value.byteValue();
+    }
+
```
chrisdutz, line 206: We now have a "byte getByte()" and a "short getBYTE()" method, I think it should only be the one or the other ... I would like to opt for the "byte getByte()". But right now I'm unsure which implications that would have.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
```diff
@@ -181,6 +205,90 @@ public short getShort() {
...
+    public boolean isInteger() {
+        return true;
+    }
+
+    @Override
+    @JsonIgnore
+    public int getInteger() {
+        return value.intValue();
+    }
+
```
chrisdutz: Instead of adding these for every type, wouldn't it make sense to have something like a "PlcIECBitStringValue", "PlcIECIntegerValue", ...?

chrisdutz: But thinking of it ... perhaps making it generic and passing in the type would help ...

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValues.java
```diff
@@ -355,6 +356,28 @@ public static PlcValue of(Map<String, PlcValue> map) {
...
+
+    public static PlcValue of(Object[] values, Class clazz) {
+        //Encode values to the type defined in clazz
+        try {
+            Constructor<?> constructor = clazz.getDeclaredConstructor(values[0].getClass());
+            if(values.length == 1) {
+                return ((PlcValue) constructor.newInstance(values[0]));
+            } else {
+                return PlcValues.of(Arrays.stream(values).map(value -> (constructorHelper(constructor, value))).collect(Collectors.toList()));
+            }
```
chrisdutz: One thing I've learnt recently, is that exceptions are extremely expensive, as creating the stack trace is expensive. is this code that is likely to produce any of "InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException" in normal execution? If yes, we should probably change this and other parts where we use similar patterns to check first instead of silently catching exceptions.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
```diff
@@ -169,6 +169,30 @@ public PlcBYTE(@JsonProperty("value") short value) {
...
+    @JsonIgnore
+    public boolean isBoolean() {
+        return true;
+    }
+
+    @Override
+    @JsonIgnore
+    public boolean getBoolean() {
+        return (value != null) && !value.equals(0);
+    }
```
chrisdutz, line 183: In general a bit-string (at least in my perception of it, represents an array of booleans. So if for example I read one byte, from the point of the API it should be a list of 8 boolean values ... I think we need to allow this somehow. Right now the only way would be to get the "byte" value and to pick that appart.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
hutcheb: There's no reason to have the getBYTE,getINT,getREAL,etc... they don't get used anymore. I added these in the previous PR so we can customize the method that gets called in the data-io template by using the case name.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValues.java
hutcheb: The exception would be raised if there wasn't a class (PlcBYTE, PlcINT, etc..) that matched the datatype that was passed. It shouldn't be raised in normal operation.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
hutcheb: I was having trouble with how we would do this. I was initially thinking that if we request to read a WORD we should parse the return value as a PlcList of PlcBOOLs instead of a PlcWORD but wasn't sure how to implement this easily.

But your idea of returning a list from a getBoolean method seems reasonable.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcBYTE.java
hutcheb: I think there a at least two use cases for these methods. To use them when parsing the value within the dataio classes. The other is by the user when they need to parse the value to a specific type. 

A generic function to be used when parsing like getValue() would be good.

For the user the getInteger, getByte, etc.. might be good as I can't think of an easy way to pass a type to the function and have it return the data type requested without just putting the existing functions within a switch statement.

PR #194, opened 2020-10-10
Add an OPC UA Server

hutcheb:


hutcheb:
It is definitely usable but somethings to improve on:-
- Security isn't implemented.
- Arrays should probably be a list of nodes in OPCUA instead of a single one.
- Discovery doesn't seem to work.

PR #196, opened 2020-10-21
Opcaua server updates

hutcheb:
- Setup user password file for OPC UA Server
- Add command line switches to handle creating users/passwords and self signed certificates.
- OPC UA server encryption now works.
- Bumped Milo to 0.52 for client and server
- getNumberOfElements method to PlcField interface.
- OPC UA server, arrays are handled now.
- df1and code-gen in the sandbox seemed to have an issue with intermittent failures when building. I think it has to do with an external dependency for slf4j.
- When parsing strings in PlcValue classes, leading spaces would cause issues.

I'll leave it open for comment for a few days before merging.


nalim2 (commented):
Looks good to me. I just inspected the code and will test it out in the afternoon.

PR #199, opened 2020-10-28
Record Evolution text adjustment

markope:
Removed a space to be able to make a PR and add Record Evolution as an adopter.


JulianFeinauer (approved):
LGTM, thank you very much @markope !

PR #200, opened 2020-11-03
PLC4X-255 Fix for Kafka Source CPU Usage

hutcheb:


plc4j/integrations/apache-kafka/src/main/java/org/apache/plc4x/kafka/Plc4xSourceTask.java
```diff
@@ -219,7 +220,13 @@ public void stop() {
             buffer.drainTo(result, numElements);
             return result;
         } else {
```
ottobackwards, line 234: Is there some way to make these configurable?

hutcheb: Well that took longer than I thought it would.

ottobackwards: I don't know if there is any logging in this area of the code, but the poll() function would seem to be a good place to put some trace logging in around what we are doing there.
I am ok with that being some kind of follow on though

ottobackwards: Actually there is almost no logging in here, not this PR's issue

ottobackwards (approved):
Thanks for the contribution and the extra work to make this configurable

PR #201, opened 2020-11-03
PLC4X-252 Support passing of PlcField in subscription builder.

splatch:
This is a follow up for earlier PLC4X-252 PR which brings same semantics to subscription requests.

@chrisdutz Given that you commented on JIRA issue I believe you are valid candidate for review.

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcSubscriptionRequest.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcSubscriptionRequest.java
@@ -18,6 +18,8 @@
  */
 package org.apache.plc4x.java.api.messages;
 
+import org.apache.plc4x.java.api.model.PlcField;
+
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 
@@ -42,6 +44,11 @@ public interface PlcSubscriptionRequest extends PlcSubscriptionFieldRequest {
         PlcSubscriptionRequest.Builder addCyclicField(String name, String fieldQuery, Duration pollingInterval);
 
         /**
+         * @see #addCyclicField(String, String, Duration)
+         */
+        PlcSubscriptionRequest.Builder addCyclicField(String name, PlcField fieldQuery, Duration pollingInterval);
+
+        /**
          * Adds a new field to the to be constructed request which should be updated as soon as
          * a value changes in the PLC.
          *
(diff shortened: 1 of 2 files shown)


chrisdutz (changes requested):
Well the reason I removed them from the API was that I couldn't see a valid scenario in which a user could create PlcField objects, as the API generally doesn't support that. If a user uses the driver module directly and directly instantiates the custom driver field types, then his code is no longer portable. So I would opt to define the PlcField based types in an interface in the SPI module. Then if a user really wants to do such things, he can explicitly use the internal API

splatch:
> If a user uses the driver module directly and directly instantiates the custom driver field types, then his code is no longer portable

That's user decision how he uses library and not ours.

Present API in 0.8 does not expose types such as `S7Connection` or `CANopenConnection` which could be used with specific field types. Even if that would be a case there is still a pooling connection which will bring yet another type to the stack.

I do instantiate fields directly in openhab bindings cause by this way I can assure that it compiles against next PLC4X releases and does not depend on changing field syntax. While PLC4X offers universal APIs the openHAB layer built on top of it provides driver and field specific configuration options. For example here you can find an updated discovery logic for one of canopen devices supported by binding:
https://github.com/ConnectorIO/connectorio-addons/commit/29c53778791796c642c9011d9adcac0e36376e39#diff-f498576d3c3ecc8a6206f7b7327ef608c0e05441b5c6ae8afcd274d5e47eebcc

Keep in mind that passing strings involves regular expressions which proved to cause minor, but still, failures. See PLC4X-56, PLC4X-194. Gluing a string with specific field syntax just to pass it to specific driver is simply redundant and subject of additional friction when migrating to future releases (when field syntax changes it goes under compiler radar).

splatch:
Ok, code is rebased to current develop and ready for further review.

sruehl:
2 Points reading your comment @splatch:

1. @chrisdutz has a valid point with moving that to the SPI
2. Regarding the "does it compile" comment: Maybe bind/use the fields in unit tests, then you would have a compile issue building your module on breaking changes and you don't pollute your productive code (You also could depend on internal artifacts with the test scope)

splatch:
The SPI is not an valid point, cause within SPI we have things such as `PlcReader`, `PassiveDiscovery` and other implementation details which are relevant from *driver* point of view. One of very few things related to use of API is `PlcProprietaryRequest` which is different from what I am asking about.
I do not consider calling specific field types in context of specific PLC driver a pollution. I believe string concatenation can do much more harm and requires much more testing.

PR #202, opened 2020-11-06
Feature/kafkasink - Add a kafka sink 

hutcheb:


ottobackwards:
not sure what is up with the conflicts?

Also, do we have tests for this?  kafka has test server classes etc, I can point you to some examples

ottobackwards:
I think the tests stuff should be a follow on, please open an issue for adding kafka integration tests

hutcheb:
Worked through the Confluent checklist as well as added tests and documentation.

- The schema for the source has remained unchanged, I have brought the sink connector schema in line with the source connector.
- I have kept the source connector names for the tags as 'fields'. So we don't break people who are using the previous version.
- Also included are fixes for the simulated driver so it can handle arrays.

chrisdutz:
As far as I know nobody is publicly using it ... :-/

So I would say: If it would be better to change things ... do it now before we're actually listed in the Confluent Hub

hutcheb:
I've been going back and froth between changing the schema for the source connector and the initial reason for it was so that we don't change the schema. Avro and the schema registry do a great job of handling changes to the schema when we add/remove fields.

Considering another case though, if we have a report by exception device that we only want to send the fields that have changed. We don't want to have to send all the fields we have specified in the config file. But only the values that have changed.

I think changing it so we have an array of [field, value] pairs would be a better option.

chrisdutz:
This is definitely a valid scenario, so I would suggest to change it.

hutcheb:
It might be a good idea to move all of the fields into their own property. this way we won't run into issues with having pre-configured names such as timestamp conflicting with field names.

chrisdutz:
What do you mean with "their own property"? If it's "their own configuration" and hereby their own kafka connect "instance" ... I strongly object. The usecase where we first used it was to collect 2600 datapoints on each PLC ... having 2600 configurations to manage just will not work.

hutcheb:
No sorry, something like:-
{
  "fields": {
                   "running": false,
                   "conveyorStop": true,
                    ......
                },
  "timestamp": 1600002002020
}

So that the user defined fields are separate to the timestamp field. This way the user can add a tag 'timestamp' and not have to worry about conflicting with the timestamp field. It also allows us to add optional fields in the future without having to worry about conflicting with peoples user defined fields.

hutcheb:
Should be in a position to merge. I'll leave it open for comments for a day or so.

plc4j/drivers/simulated/src/main/java/org/apache/plc4x/java/simulated/connection/SimulatedDevice.java
```diff
@@ -97,12 +97,12 @@ public void set(SimulatedField field, PlcValue value) {
...
                     default:
                         try {
-                            DataItemIO.staticSerialize(value, field.getPlcDataType(), 1, false);
+                            DataItemIO.staticSerialize(value, field.getPlcDataType(), field.getNumberOfElements(), false);
                         } catch (ParseException e) {
                             System.out.printf("Write failed");
                         }
                 }
-                System.out.printf("TEST PLC RANDOM [%s]: %s%n", field.getName(), value.getString());
+                System.out.printf("TEST PLC RANDOM [%s]: %s%n", field.getName(), value.toString());
```
chrisdutz: We probably shouldn't be using System.out.prlint stuff

plc4j/integrations/apache-kafka/README.md
```diff
@@ -60,7 +66,7 @@ In order to start a Kafka Connect system the following steps have to be performe
         
         bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic test --from-beginning
 
-### Start a Kafka Connect Worker (Standalone)
+### Start a Kafka Source Connect Worker (Standalone)
```
chrisdutz: The name of the product is called "Kafka Connect" ... as this is a Source for Kafka Connect, I'd probably name it "Kafka Connect Source Worker"

plc4j/integrations/apache-kafka/README.md
```diff
@@ -76,7 +82,7 @@ If you want to debug the connector, be sure to set some environment variables be
 
 In this case the startup will suspend till an IDE is connected via a remote-debugging session.
 
-### Start Kafka Connect Worker (Distributed Mode)
+### Start Kafka Source Connect Worker (Distributed Mode)
```
chrisdutz: Same as above

plc4j/integrations/apache-kafka/README.md
```diff
@@ -95,3 +101,52 @@ The configuration of the Connectors is then provided via REST interface:
     curl -X POST -H "Content-Type: application/json" --data '{"name": "plc-source-test", "config": {"connector.class":"org.apache.plc4x.kafka.Plc4xSourceConnector", 
     // TODO: Continue here ...
     "tasks.max":"1", "file":"test.sink.txt", "topics":"connect-test" }}' http://localhost:8083/connectors
+
+
+### Start a Kafka Sink Connect Worker (Standalone)
```
chrisdutz: Here, the thing should be named "Kafka Connect Sink Worker"

plc4j/integrations/apache-kafka/pom.xml
```diff
@@ -33,8 +33,17 @@
 
   <properties>
     <kafka.version>2.5.0</kafka.version>
+    <kafka.connect.maven.plugin.version>0.11.3</kafka.connect.maven.plugin.version>
   </properties>
 
+  <repositories>
+    <repository>
```
chrisdutz, line 40: As this is used to download the plugin, we need to also specify a "pluginRepository" (Just a duplicate with different element name) ... otherwise the build will fail for people who haven't downloaded the plugin before.

chrisdutz (changes requested):
I didn't look into the details of the code-changes themselves as I have almost no experience with Kafka and Kafka Connect ... just responded to some build-related and naming issues.

Especially the pluiginRepository issue should be addressed before merging.

plc4j/integrations/apache-kafka/README.md
```diff
@@ -1,82 +1,190 @@
...
 
 ### Start a Kafka Broker
 
 1) Open 4 console windows and change directory into that directory
-2) Start Zookeeper: 
-        
-        bin/zookeeper-server-start.sh config/zookeeper.properties 
+2) Start Zookeeper:
+
+        bin/zookeeper-server-start.sh config/zookeeper.properties
```
chrisdutz, line 64: I even heard that new versions of Kafka work without Zookeeper ... we should check this and perhaps update this section later on

chrisdutz (approved):
All looks good to me (With the actual code changes I 100% have faith in what you did) ... thanks for your hard work

PR #203, opened 2020-11-08
PLC4X-257 Support timeout cases for late conversations

splatch:
I composed two commits - one is fix for missing `onTimeout` calls, second is one which helped me finding bug.

I know we got couple of extra lines with extra classes in place of lambdas, however stack trace reading is much easier now. Given that pretty much any conversation involve multiple `only` statements it will impact every protocol.


JulianFeinauer (approved):
Thanks for taking care yourself. You did it exactly the way it should have been done already and even cleaner then I would ahve hacked it. Feel free to merge!

splatch:
I am closing this one as it is now separated in PLC4X-258. Timeout fixes will go under PLC4X-207.

PR #210, opened 2020-12-12
Bug/driver opcua read arrays [PLC4X-202]

hutcheb:
Add support for data types specifier and add support for writing arrays.

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaField.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaField.java
@@ -30,7 +30,7 @@ import java.util.regex.Pattern;
  */
 public class OpcuaField implements PlcField {
 
-    public static final Pattern ADDRESS_PATTERN = Pattern.compile("^ns=(?<namespace>\\d+);(?<identifierType>[isgb])=((?<identifier>.+))?");
+    public static final Pattern ADDRESS_PATTERN = Pattern.compile("^ns=(?<namespace>\\d+);(?<identifierType>[isgb])=((?<identifier>[^:]+))?(:(?<datatype>[a-zA-Z_]+))?");
 
     private final OpcuaIdentifierType identifierType;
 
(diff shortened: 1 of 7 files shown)


plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaField.java
```diff
@@ -30,30 +30,34 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+
+    protected OpcuaField(int namespace, OpcuaIdentifierType identifierType, String identifier, String dataType) {
         this.namespace = namespace;
         this.identifier = identifier;
         this.identifierType = identifierType;
         if (this.identifier == null || this.namespace < 0) {
             throw new IllegalArgumentException("Identifier can not be null or Namespace can not be lower then 0.");
         }
+        this.dataType = dataType != null ? dataType.toUpperCase() : null;
     }
```
ottobackwards, line 52: Is there a limited set of datatype?  or can it be anything?  If it is limited, shouldn't this be an enumeration?

hutcheb: I don't see any reason why we can't change this to an enum. To keep it the same as the others I'll add an mspec for opc ua for the datatype enum.

chrisdutz: Problem with enums are that they are not extendable. So if we have a driver that generally supports the normal IEC types, but has more, we couldn't reference them

hutcheb: At the moment I have to explicitly map between PlcValue types and OPC UA data types when writing data. There's a bunch of if statements in there checking what datatype it is. 

If people want to extend it and add datatypes they would have to change this logic to take care of the new data type anyway. If we used an enum we could possibly (I'd have to check to see if we would end up using a class lookup) include the OPC UA data type to map it to. This way if people wanted to extend it they would only need to update the mspec as well as add whatever PlcValue class they want.

hutcheb: Without using class lookups, I'm not sure if there is a better way to handle the PLC4X->Milo data type mapping within plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java

Having a project goal of being able to easily provide a way to extend drivers with custom data types, adding an enum for the data types wouldn't be towards that.

Having a function similar to <T> getObject() within PlcValue instead of Object getObject() might be one way to improve the logic but will leave that open for a future work.

ottobackwards: ok.  my concern was around the lack of checking for validity.  enums would handle that.

PR #213, opened 2020-12-17
PLC4X-265 Support for hex mapping of byte/byte[] values in XML test

splatch:
There are small updates in EIP stuff which shouldn't break anything.

### DIFF
--- a/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/readwrite/field/EipField.java
+++ b/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/readwrite/field/EipField.java
@@ -20,11 +20,14 @@ package org.apache.plc4x.java.eip.readwrite.field;
 
 import org.apache.plc4x.java.api.model.PlcField;
 import org.apache.plc4x.java.eip.readwrite.types.CIPDataTypeCode;
+import org.apache.plc4x.java.spi.utils.XmlSerializable;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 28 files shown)


chrisdutz:
Hmm ok ... so you extended the API model classes with explicit serialization code ... would it be possible to have this sort of outside of the API classes? If included in the API, I would prefer the default Java XML DOM instead of Dom4j ... I Like the DOM4J api more than the default, but I wouldn't want to force that on the users.

splatch:
> I would prefer the default Java XML DOM instead of Dom4j

Which place do you mean - plc field implementation or test framework?

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/readwrite/field/EipField.java
```diff
@@ -135,4 +138,27 @@ public String getPlcDataType() {
...
+        messageElement.appendChild(indexElement);
+
+        Element subIndexElement = doc.createElement("elementNb");
+        subIndexElement.appendChild(doc.createTextNode(Integer.toString(elementNb)));
+        messageElement.appendChild(subIndexElement);
+
+        Element dataType = doc.createElement("defaultJavaType");
+        dataType.appendChild(doc.createTextNode((type == null ? Object.class : getDefaultJavaType()).getName()));
+        messageElement.appendChild(dataType);
+    }
```
chrisdutz, line 163: I mean this stuff ..

splatch: This is regular implementation of `XmlSerializable` interface from plc4x spi.utils package. It is based on regular dom api just like other places.

Overall issue is how to make this mapping automatic in xstream, am I right?

PR #214, opened 2020-12-19
Add enum's for opc ua driver in mspec

hutcheb:
Fix for connection string optional fields tcp and params.


plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaField.java
```diff
@@ -68,12 +69,14 @@ public static OpcuaField of(String address) {
         String identifier = matcher.group("identifier");
 
         String identifierTypeString = matcher.group("identifierType");
-        OpcuaIdentifierType identifierType = OpcuaIdentifierType.fromString(identifierTypeString);
+        OpcuaIdentifierType identifierType = OpcuaIdentifierType.enumForValue(identifierTypeString);
```
ottobackwards, line 73: What happens if it is an invalid value?

ottobackwards (commented):
One question inline.  Also, are there any tests for the address parsing?

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaField.java
hutcheb: A null pointer exception is eventually raised, might be worth adding a contains key to the enum template as it looks like it is a concern with all of the Field classes.

ottobackwards (approved):

PR #219, opened 2021-01-07
Feature/plc4py

ottlukas:
This Pull request aims to generate python code using the code generation utils. to create the basis for developing plc drivers in python. Using mspec and code generation


ottlukas:
https://cwiki.apache.org/confluence/display/PLC4X/Python+-+Built+Utils

ottobackwards:
Please add a description of what you are trying to do in this PR and any verification instructions.

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.poetry]
+name = "plc4x"
+version = "0.1.0"
+description = ""
```
ottobackwards: I'm not sure if this should be individuals or just plc4x project or something.  @chrisdutz ?  if individuals, add me :)

pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+authors = ["ottlukas <ott.lukas.14@gmail.com>"]
```
chrisdutz: This should better be "Apache PLC4X PMC"

ottlukas: yes ... strange in PyCharm it looks like this:
image
But when I push to repo it does not...any ideas how to change this (seems to be IDE related)

pyproject.toml
chrisdutz: In general we should never put individuals listed anywhere ... it's part of the legal shield the ASF can protect us with.

sandbox/plc4py/pyproject.toml
```diff
@@ -24,6 +24,7 @@ authors = ["Apache PLC4X <dev@plc4x.apache.org>"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
+numpy = "^1.19"
 
 [tool.poetry.dev-dependencies]
 pytest = "^5.2"
```
chrisdutz: Wouldn't it be better to put explicit versions in here and not version ranges? (Or is this more like our prerequisite check in PLC4X where we check at least a given version is installed?)

ottlukas: 1) Already deleted numpy
2) We will follow PEP-517 https://www.python.org/dev/peps/pep-0517/ and define which minimum version of a module is required for the building.

chrisdutz: But doesn't this make things not reproducible?

pyproject.toml
ottobackwards: we need to tell the folks at commons that then yikes

sruehl:
@ottlukas maybe it might be a good idea to just rebase this branch onto develop as its really hard to see whats contained within this branch. Should be fairly easy with the interactive-rebase feature of intellij.
Usually you don't rebase pushed branches but as this branch is contained within you personal fork we should be good.
WDYT?

ottlukas:
> @ottlukas maybe it might be a good idea to just rebase this branch onto develop as its really hard to see whats contained within this branch. Should be fairly easy with the interactive-rebase feature of intellij.
> Usually you don't rebase pushed branches but as this branch is contained within you personal fork we should be good.
> WDYT?

Yes just rebase is fine with me

sruehl:
I merge this now into a temporary branch and try to cleanup the history there,

PR #220, opened 2021-01-08
PLC4Py maven integration

vemmert:


ottobackwards:
we are using poetry, you don't need the setup tools

poetry install && poetry build

ottobackwards:
As per our discussion, we need to do more to make this work in the scenarios:
new user not using mvn
new user using mvn

subsequent mvn builds

with regards to virtual env, setup, dependency installs.

We can land this, i'll create a follow on for making that stuff work

ottobackwards (approved):
follow on required

PR #221, opened 2021-01-10
Bug/close connection

hutcheb:
Added call to disconnect when closing a connection.
This allows the connection to wait for the protocol to cleanly disconnect before closing the channel.


ottobackwards:
+1,  I think this is correct. 
Two points though.
- It is not good that there  are no tests for that, but I don't think that is this PR's responsibility.
- This kind of flow if really difficult to grok and troubleshoot.  The complete lack of logging here doesn't help things.  But again that is beyond this PR's responsibility
-

hutcheb:
I agree tests for this class should be implemented, I also agree that it can be handled as a separate issue.

I've added a small amount of logging so it can be traced from Call DisConnectEvent->OnDisconnect->Call DisconnectedEvent->Return.

PR #223, opened 2021-01-16
Updated Notice and License files.

hutcheb:
To take into account for derivative works licenses under EPL 2.0


pom.xml
```diff
@@ -679,10 +679,13 @@
...
             <!--
-              Explicitly exclude this file as I was unable
-              to add a new license to the configuration.
+              Explicitly exclude these files as I was unable
+              to add a new licenses to the configuration.
             -->
             <exclude>**/Expression.g4</exclude>
+            <exclude>**/KeyStoreLoader.java</exclude>
+            <exclude>**/OPCUAServer.java</exclude>
+            <exclude>**/Plc4xNamespace.java</exclude>
```
chrisdutz, line 688: Actually you shouldn't need these exclusions. The rat tool should detect the EPL header. But it doesn't harm either.

chrisdutz (approved):
Looks good to me ... we probably still need to have stuff added to the kafka connect module itself, but I'll help with this.

PR #225, opened 2021-01-20
Add encryption handler for OPC UA - Minor fix for Kafka Connector

hutcheb:
I've added the Encryption handler for the OPC UA server and re-wrote some of the logic. I have worked on a add just what we need to the logic, we can build from here.

If an error occurs, in some cases,  while polling for the OPC UA server all the tags will stop. This has been fixed so that only the failed tag stops and shows a bad value.

Also a fix based on Rankesh's comment about adding jitter to the poll return value for the Kafka Source Connector. This distributes some of the load on the Kafka servers when a lot of connectors are all started at the same time.


plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xNamespace.java
```diff
@@ -1,11 +1,20 @@
 /*
- * Copyright (c) 2019 the Eclipse Milo Authors
```
nalim2, line 2: i am no expert on this license topic. But is this ok in that way? That we have the small Milo notice in the "Notice" file and change the license?

chrisdutz: Unfortunately not ... EPL is category B ... we can use (in form of use libaries) but we can't include that code. ... not even just a bit. If it's just a bit, it would be better to re-write it.

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
```diff
@@ -115,9 +71,27 @@
...
     public Plc4xCommunication () {
+
+    }
+
+    @Override
+    protected void onStartup() {
         driverManager = new PooledPlcDriverManager();
     }
 
+    @Override
```
ottobackwards, line 83: are you hiding the base class onShutdown?
If so, I would comment this.

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
```diff
@@ -195,93 +169,110 @@ public static NodeId getNodeId(String plcValue) {
     }
 
     public DataValue getValue(AttributeFilterContext.GetAttributeContext ctx, String tag, String connectionString) {
-        PlcConnection connection = null;
         DataValue resp = new DataValue(new Variant(null), StatusCode.BAD);
```
ottobackwards: Maybe we should have a static constant,  Datavalue resp = DataValue.BAD_STATUS

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
```diff
@@ -195,93 +169,110 @@ public static NodeId getNodeId(String plcValue) {
...
-        } catch (PlcConnectionException e) {
-            logger.error("Failed to connect to device, error raised - " + e);
-            failedConnectionList.put(connectionString, System.currentTimeMillis());
-            return resp;
-        }
-
-        if (!connection.getMetadata().canRead()) {
-            logger.error("This connection doesn't support reading.");
+            //Try to connect to PLC
             try {
```
ottobackwards, line 187: Question, as this is a PLC4x client that exposes things as OPCUA, wouldn't we want the client to be using the pool?

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xNamespace.java
```diff
@@ -229,6 +183,11 @@ private void addDynamicNodes(UaFolderNode rootNode, DeviceConfiguration c) {
     public void onDataItemsCreated(List<DataItem> dataItems) {
         for (DataItem item : dataItems) {
             plc4xServer.addField(item);
+
+            if (plc4xServer.getDriverManager() == null) {
+                plc4xServer.removeField(item);
```
ottobackwards, line 188: ok, we are using pooled, sorry

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/context/Encryption.java
```diff
@@ -0,0 +1,122 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.math.BigInteger;
+import java.net.InetAddress;
+import java.security.*;
+import java.security.cert.X509Certificate;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.Locale;
+
```
ottobackwards: Can we rename this for the context?  Encryption seems a little too generic

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xNamespace.java
hutcheb: The EPL headers for the Plc4xNamespace and OPCUAServer files were added in as I wasn't sure how much of the code reflected the Milo example.
I am comfortable that there is no code that needs to be licensed under EPL remaining.

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
hutcheb: It overrides an abstract method, I don't really have much to do on shutdown. I've added a comment to indicate that.

hutcheb:
Otto, all good ideas, I have updated the code.

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
```diff
@@ -65,6 +65,7 @@
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private final Integer DEFAULT_TIMEOUT = 1000000;
     private final Integer DEFAULT_RETRY_BACKOFF = 5000;
+    private final DataValue BAD_RESPONSE = new DataValue(new Variant(null), StatusCode.BAD);
```
ottobackwards, line 68: should this be static?

plc4j/integrations/opcua-server/src/main/java/org/apache/plc4x/java/opcuaserver/backend/Plc4xCommunication.java
ottobackwards: ah, IC,  you can't tell if it is abstract base or interface, always catches me.  thanks

ottobackwards (approved):
Just a question about making the BAD_RESPONSE static.
But that is a nit.

PR #226, opened 2021-02-06
udp protocols  format send and resive 

problemSolvingProgramming:
How can I directly access the data I receive from the DCS device via Deltav? I know it is through opc but I do not want to use opc.
I have the HMI that I received through the following link    https://github.com/GavinYellow/SharpSCADA , I design my doctoral research project, but unfortunately I do not know how I can access the data directly through ifix, except through opc. If you can tell me the protocol that Deltawi communicates with ifix, thank you.
I do not want to use it because Deltavi uses the old version of ifix and it also has some weaknesses. If possible, please tell me how I can access Delta data directly using the following SCADA code  https://github.com/GavinYellow/SharpSCADA


hutcheb:
Hi doctorwho-123,
You seem to have created a Pull Request to ask a question, you would be better off subscribing to the mailing list dev@plc4x.apache.org, using the address dev-subscribe@plc4x.apache.org. And ask you question there.

The PLC4X DeltaV driver I believe is only implemented in Java, SharpSCADA seems to use C#, we do have a framework for C# however I don't believe there are any drivers implemented yet.

Kind Regards

Ben

PR #227, opened 2021-03-05
Fix for opcua subscription

hutcheb:


plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
```diff
@@ -384,8 +385,8 @@ public void close() throws Exception {
         CompletableFuture<PlcSubscriptionResponse> future = CompletableFuture.supplyAsync(() -> {
             Map<String, ResponseItem<PlcSubscriptionHandle>> responseItems = new HashMap<>();
             for (String fieldName : subscriptionRequest.getFieldNames()) {
-                final PlcSubscriptionField subscriptionField = subscriptionRequest.getField(fieldName);
-                final OpcuaField field = (OpcuaField) Objects.requireNonNull(subscriptionField);
+                final DefaultPlcSubscriptionField subscriptionField = (DefaultPlcSubscriptionField) subscriptionRequest.getField(fieldName);
```
splatch, line 388: Shouldn't we be fine with just `PlcSubscriptionFiled` alone?

hutcheb: We are using the DefaultPlcSubscriptionRequest class which creates a DefaultPlcSubscriptionField containing the OpcuaField. This needs to be cast to DefaultPcSubscriptionField so we can get the PlcField from it. It would be good if the DefaultPlcSubscriptionRequest created a PlcSubscriptionField based off that returned by the PlcDriver though.

At the moment the OpcuaField doesn't need to implement the PlcSubscriptionField.

nalim2:
Seems good to me

nalim2 (approved):

PR #229, opened 2021-03-17
hello_world_s7 (C) now working

thomas169:
hello-world-s7 now working when passed connection string with port eg. `s7:tcp://0.0.0.0:102` I don't think it possible that it was working before... ?

`spi/src/system.c`: plugged mem leaks, removed unneeded mallocs (ie. mallocs in both `spi/src/system.c` and `spi/src/connection.c`, needed only in 1 of these locations), fixed compiler warnings on `hurz` cast and de-globalized `hurz` (maybe globalness was intended?). Rewrote the connection string parser to use tokenisation rather than walking pointer, which is IMO much clearer.

`hello_world_s7.c` beautified some `printf`s and allowed optional cli arguments for setting the test connection string. 

`transport_tcp.c` fixed compile warning on `printf` without const format specifier.

### DIFF
--- a/sandbox/plc4c/examples/hello-world-s7/src/hello_world_s7.c
+++ b/sandbox/plc4c/examples/hello-world-s7/src/hello_world_s7.c
@@ -21,7 +21,6 @@
 #include <plc4c/transport_tcp.h>
 #include <stdio.h>
 #include <string.h>
-#include <unistd.h>
 
 
 #include "../../../spi/include/plc4c/spi/types_private.h"
@@ -151,8 +150,6 @@ int main(int argc, char** argv) {
         } else if (plc4c_connection_has_error(connection)) {
           printf("FAILED\n");
           return -1;
-        } else {
-          usleep(1000*100);
         }
         break;
       }
--- a/sandbox/plc4c/examples/hello-world-s7/src/hello_world_s7.c
+++ b/sandbox/plc4c/examples/hello-world-s7/src/hello_world_s7.c
@@ -21,8 +21,12 @@
 #include <plc4c/transport_tcp.h>
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
+
 
 #include "../../../spi/include/plc4c/spi/types_private.h"
+
+#define DEFAULT_CONNECTION_TEST_STRING "s7:tcp://192.168.23.30:20"
 
 int numOpenConnections = 0;
 
@@ -58,12 +62,19 @@ typedef enum plc4c_connection_state_t plc4c_connection_state;
 //#pragma clang diagnostic push
 //#pragma ide diagnostic ignored "hicpp-multiway-paths-covered"
 
-int main() {
+int main(int argc, char** argv) {
+  
+  char* connection_test_string;
   bool loop = true;
   plc4c_system *system = NULL;
   plc4c_connection *connection = NULL;
   plc4c_read_request *read_request = NULL;
   plc4c_read_request_execution *read_request_execution = NULL;
+
+  if (argc == 1)
+    connection_test_string = DEFAULT_CONNECTION_TEST_STRING;
+  else
+    connection_test_string = argv[1];
 
   // Create a new uninitialized plc4c_system
   printf("Creating new PLC4C System (Initializing inner data-structures) ... ");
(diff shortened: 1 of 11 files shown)


sandbox/plc4c/examples/hello-world-s7/src/hello_world_s7.c
```diff
@@ -21,9 +21,13 @@
 #include <plc4c/transport_tcp.h>
 #include <stdio.h>
 #include <string.h>
```
ottobackwards, line 23: Do we need windows support?  I think we need some ifdef magic here no?

sandbox/plc4c/spi/src/system.c
```diff
@@ -204,146 +204,71 @@ void plc4c_system_shutdown(plc4c_system *system) {}
...
-          strncpy(parameters, start_segment, (i - start_segment_index) + 1);
-        }
-      }
-    }
-  }
-  if (num_colons == 0) {
-    return INVALID_CONNECTION_STRING;
-  }
+  // Make a local copy so not to mess up oringal argument, needed later?
+  strcpy(connection_to_tokenize, connection_string);
```
ottobackwards: spelling

sandbox/plc4c/spi/src/system.c
```diff
@@ -204,146 +204,71 @@ void plc4c_system_shutdown(plc4c_system *system) {}
...
+  new_connection = malloc(sizeof(plc4c_connection));
   plc4c_connection_initialize(new_connection);
   plc4c_connection_set_connection_string(new_connection, connection_string);
-  plc4c_connection_set_protocol_code(new_connection, protocol_code);
-  plc4c_connection_set_transport_code(new_connection, transport_code);
-  plc4c_connection_set_transport_connect_information(
-      new_connection, transport_connect_information);
-  plc4c_connection_set_parameters(new_connection, parameters);
 
+  // Get the protocol code (1st item, ':' delimited, required)
```
ottobackwards: windows again strtok_s

sandbox/plc4c/spi/src/system.c
```diff
@@ -204,146 +204,71 @@ void plc4c_system_shutdown(plc4c_system *system) {}
...
+  connection_token = strtok_r(connection_to_tokenize, ":", &connection_string_pos);
```
ottobackwards: strtok_r can return NULL, this needs to check no?

sandbox/plc4c/spi/src/system.c
```diff
@@ -204,146 +204,71 @@ void plc4c_system_shutdown(plc4c_system *system) {}
...
+  if (mode == SEARCHING_FOR_COLONS) {
+    connection_token = strtok_r(connection_string_pos, ":", &connection_string_pos);
+    plc4c_connection_set_transport_code(new_connection, connection_token);
+    if (strncmp(connection_string_pos, "//", 2)) 
+      return INVALID_CONNECTION_STRING;
+  }
+
+  // Skip over the '//' we have now asserted MUST exist, and search for
+  // connection parameters (last item, '?' delimited, optional).
+  connection_string_pos += 2;
```
ottobackwards: again NULL checks

ottobackwards:
Thanks for the contribution!  Just a couple of questions.
Also, we have to find some way to test the connection string stuff

sandbox/plc4c/spi/src/system.c
thomas169: plc4c_connection_set_transport_code and similar "set" functions check for null hence I didn't, but actually yes if it reutns null subsquency opeations should not be performed as the last example shows

hutcheb:
Hi Thomas,

Just a little confused why this was closed. Closing a PR basically just deletes the PR without having it merged, was that your intention?

Ben

thomas169:
As a git/github noobie I didn't realize that push's to my forked repo automatically went into this PR. Yesterday, I spent the full day getting a write->read loopback test working (fast s7:tcp IO is all I really care about for now) . And commited this, but alas unintentionally to this PR also. So I closed it.

Will make a new PR soon, but the changes are now quite unwieldy (large) for review.  Anyway I guess the good news is that I can write a value to a DB and read it back correctly.

hutcheb:
Yeah that got me at first as well, but I eventually realized that it doesn’t really matter. If you leave anything unintended in a commit you always make another commit to fix it.

GitHub has a decent review section in that it will only show the differences between the code at the last commit and the develop branch. If you add code and then remove it won’t show up in the review.

If you aren’t completely happy with the status of the PR you can always change it to a draft PR so it doesn’t get merged by accident. But using the same PR will keep all the comments and reviews in the one place.

You will also probably need to fill out an agreement with The Apache Foundation (ICLA or CCLA) before  something gets merged. I’ll dig up a link that explains it.

hutcheb:
https://www.apache.org/licenses/contributor-agreements.html

This explains a bit about what the ICLA/CCLA are and how to submit one.

Welcome to the PLC4X project by the way. :)

PR #235, opened 2021-04-10
feat: Add read/write support for string types within the Java dataio classes

hutcheb:
Following on from adding variable string length to mspec it was now possible to more easily add read/write support for strings.
Also removed a lot of the IEC61131 string indentifiers with opcua and Simulated protocols.


sruehl:
LGTM

PR #237, opened 2021-04-11
Fix/modbus string

hutcheb:
Java implementation for modbus strings using the format 400001:STRING(6)[3] 
(6) is the string length
[3] is how many strings to read.

C and Go would need to be updated.


chrisdutz:
Hi Ben ... what's the status of this PR?
And do I understand it correctly, that this applies the same logic as with S7 for Modbus?

hutcheb:
Yes it applies the same principle as the S7 String logic. It ended up being more work than I was anticipating, but for Java it works. I haven't looked at C and Go though which is why it is still here.

I probably won't end up updating C and Go as there doesn't seem to be interest in it.

chrisdutz:
I think it's ok to not have 100% parity between all ... after all the Go KNX can do way more than the Java KNX ...

chrisdutz:
You could just create two Jira issues for Go and C. Then just merge it (and add a comment to the RELEASE_NOTES) and be done with it ;-)

hutcheb:
There's a few changes in the mspec which causes a few issues in the other languages. I think at the moment we should at least be able to read strings in the other languages, we'd probably lose that if I hacked it together to be able to merge it.

chrisdutz:
I think I never really implemented String reading in go (or in C or in Java) ... so if the code generation produces something that compiles, I woudln't have any pain merging it. Do people read Strings in Modbus anyway? It definitely sounds totally inefficient.

hutcheb:
That's the other thing, I can guarantee that there's not going to be a standard way of representing strings across PLCs. 
I can't see it as being any more inefficient than reading other data types, you can read 240 odd bytes per request.

chrisdutz:
@hutcheb Is this branch still valid? If yes, we would need to rebase it and port the changes back .

PR #238, opened 2021-04-11
Remove merge button when merging PR's, this forces the use of squash

hutcheb:
Following on from Chris's comment about using a standardized commit comments, What are people thoughts on removing the merge button infavour of squash when merging PR's?


sruehl:
👍 we could try and in case we need the button back we can always re-enable.

PR #239, opened 2021-04-13
Update S7Field.java

Meng5:
fix S7 Driver read type of String with the address copied directly out of TIA portal bug.


sruehl:
thanks for the PR

PR #243, opened 2021-04-29
plc4go: fixed passing parameters incorrectly, resulting in a null pointer

hongjinlin:
When read 000019:BOOL throw "recovered from runtime error: invalid memory address or nil pointer dereference"
{"level":"trace","time":"2021-04-30T19:05:40+08:00","message":"convert response to ADU"}
{"level":"trace","time":"2021-04-30T19:05:40+08:00","message":"convert response to PLC4X response"}
{"level":"error","time":"2021-04-30T19:05:40+08:00","message":"recovered from runtime error: invalid memory address or nil pointer dereference"}


sruehl:
thx for the PR 👍

PR #244, opened 2021-05-05
plc4c: memory plumbing

thomas169:
started resolving memory issues pulled up by valgrind. 
Currently all IO leaks so doing IO in while loop crashes after not very long.


thomas169:
I have not verified changes to template result in what is needed (commend mallocs for optional items), but quite sure it will be fine

thomas169:
can now connect and disconnect wilthout leaks

thomas169:
IO now doesn't leak... so says valgrind. 
Lots of changes in PR (I know I can commit now direclty) but would quite like to keep memory alloc changes in this PR.

chrisdutz:
What's the status on this?

thomas169:
It's finished.

chrisdutz:
Aaah ok... Seems there are some conflicts... I'll have a look

chrisdutz:
Ok ... so I just pulled your latest changes and updated the branch with the latest changes from develop ... it seems you commented out some code in the code generation for optional fields. It seems this is now causing segmentation faults in the testsuite ...

thomas169:
Okay the aim is to comment out optional field allocs in copt_packet.c and s7_message.c as they get double alloc'd. 

If other optional field allocs need to stay then I guess some slightly bigger code-gen changes will be needed.

chrisdutz:
So I just pushed the merged branches to "feature/plc4x-memory-cleanup" in the official plc4x repo. As you are now a full committer, I think we should continue there (makes the review, updating, merging simpler)

chrisdutz:
Well I think only the parser can know if it needs to alloc memory for the optional field. But code like this will always produce segfaults:

`      
//dataUnitReferenceNumber = malloc(sizeof(uint8_t));
//if(dataUnitReferenceNumber == NULL) {
//  return NO_MEMORY;
//}
*dataUnitReferenceNumber = 0;
`

chrisdutz:
But having a look at the code ... I think if we move the malloc to the "simple" path, this should not double-allocate ... correct?

thomas169:
Yes that is unintended collateral damage as it not in `s7_message.c` or` copt_paramtmer.c`. 

Ill have a look at what the code-gen outputs with changes to templates (I did not before, hence all the explosions) sometime this week.

chrisdutz:
I just pushed a change to the new branch which got rid of the segfault while not double allocating memory for the complex-type case but only for optional fields with simple types ... please have a look if that's a good compromise.

chrisdutz:
Sorry for the delay by the way ... I was consumed with a lot of "work stuff" (TM) ... but really looking forward to integrating your changes. As soon as that's done, I really would love moving the PLC4C directory out of the sandbox and into the root of the project (Making it something we consider somewhat mature)

PR #253, opened 2021-05-28
Feature/native opua client

hutcheb:
Replaced the Opcua driver with a native one

Fixed mspec issue around casting numeric values as string parameters to discriminated types. They now use .valueOf instead of trying to cast it to a String.

Added a discover callback for Java connections to be used once the discovery process has completed.


hutcheb:
I have to work on message chunking tomorrow, but there shouldn't be any major changes. I should be able to do some minor changes for the next few days but afterwards I'll have to sign off.

PR #259, opened 2021-07-02
PLC4X-304 Improve status code mappings for OPC UA.

splatch:
Improved mapping for error codes. Turns out that we can detect security error using Milo apis. More over status code comparison should use `.isGood()` methods cause these are results of comparison at bit mask level.

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java
@@ -513,12 +513,11 @@ public class OpcuaTcpPlcConnection extends BaseOpcuaPlcConnection {
             for (int counter = 0; counter < readValueIds.size(); counter++) {
                 PlcResponseCode resultCode = PlcResponseCode.OK;
                 PlcValue stringItem = null;
-                if (readValues == null || readValues.size() <= counter ||
-                    !readValues.get(counter).getStatusCode().equals(StatusCode.GOOD)) {
+                if (readValues == null || readValues.size() <= counter || readValues.get(counter) == null) {
                     resultCode = PlcResponseCode.NOT_FOUND;
                 } else {
+                    resultCode = mapStatusCode(readValues.get(counter).getStatusCode());
                     stringItem = encodePlcValue(readValues.get(counter));
-
                 }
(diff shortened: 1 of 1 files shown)


GlennZeng:
Thank you for your dedication! I want to ask if there is an opcua code case. I want to refer to it. I searched a lot and couldn't find it. I saw that there seems to be no place to enter the opcua account password in your connection code. What's the matter?

splatch:
Hey Glen,
Currently the high level API of PLC4X does specify a concept of credentials, but these as far as I know are not being used by any of drivers. Even for OPC UA where there is a support for various kinds of authentication it is not being utilized.

Have a look on https://github.com/apache/plc4x/blob/v0.8.0/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/connection/OpcuaTcpPlcConnection.java#L343L349 - this is place where Eclipse Milo client is being initialized. It should rely on non-anonymous identity provider.
I believe @hutcheb will know best how future handling of this will look a like.

beyondyinjl2:
Is there a golang opcua driver?

ottlukas:
> Is there a golang opcua driver?

Not yet

GlennZeng:
Thanks for your reply! I would also like to ask, what protocols does golang currently support? What protocols does Java temporarily support?

ottlukas:
> Thanks for your reply! I would also like to ask, what protocols does golang currently support? What protocols does Java temporarily support?

https://plc4x.apache.org/users/protocols/index.html see our Table here -> Java supports most of the protocols -> Golang -> KNX, S7 and Modbus . C -> S7

hutcheb:
> Thank you for your dedication! I want to ask if there is an opcua code case. I want to refer to it. I searched a lot and couldn't find it. I saw that there seems to be no place to enter the opcua account password in your connection code. What's the matter?

I have merged a major update for the OPCUA protocol into develop, it should allow you to specify a username and password as part of the connection string e.g. opcua:tcp://127.0.0.1:4840/node?username=fred;password=fredspassword

For the status codes, this should have all of them available in opcua,
https://github.com/OPCFoundation/UA-Nodeset/blob/v1.04/Schema/StatusCode.csv
I'll double check how it is now handled, but if you could take a look at the update and comment it woulld be very helpful.

splatch:
Given that we got brand new and fresh OPC UA I'll close this PR. Above code can be used as an patch for 0.8 release.

PR #263, opened 2021-07-29
 PLC4X-307 Add support for custom generator options / customized packages

splatch:
These two commits are follow up to build tools changes. First commit makes sure generators are fine and can compile, other one opens test framework for use of customized output packages.


sruehl (approved):
bit stringly typed so maybe we can add some sanity checks that protocolName, and output flavor is present on getMessageIOType but other than that LGTM

chrisdutz (approved):

splatch:
> bit stringly typed so maybe we can add some sanity checks that protocolName, and output flavor is present on getMessageIOType but other than that LGTM

I agree, it is indeed yet another string trap. I'd wish we could migrate to typed configuration and minimize amount of arguments passed over test framework (language, protocolName, driverName). Yet for that I believe we need to book some exclusive time reserved only for test framework.

PR #265, opened 2021-08-13
Subscriptions for ADS in PLC4J

rmeister:
Hi,

I've got working subscriptions for ADS in PLC4J. As Łukasz pointed out, it wasn't that much effort. 
What I'm not sure about is how to add tests. Did some manual tests by tweaking the subscription example and it looked fine so far (unsubscribe working as well).

Happy to hear your comments.

Cheers,
Richard


splatch:
Hey @rmeister,
Thank you for your awesome work. After all looking at amount of code is not that short. :-)

If you would wish to test the logic and make sure that it works with certain kind of packets coming from ADS then the best way would be to utilize driver test suite. This is a XML based descriptor of API operations to executed by PLC4X driver, expected messages sent to transport, messages answered by end device as well as messages sent by PLC. Please have a look at canopen sample:
https://github.com/apache/plc4x/blob/feature/can-stack-tunning/protocols/canopen/src/test/resources/protocols/canopen/CANOpenDriverSDOIT.xml
(note this is in branch, it is intended to be merged into develop)
There you can find a test which does several wire operations to confirm that segmentation works fine. Similar thing could be done for ADS driver. Please don't be afraid of lengthily XML. I believe it can be generated with little helper programs pushing data frame into `WriteBufferXmlBased`.

Cheers,
Łukasz

rmeister:
Hey @splatch 
thank you for the explanation and hints.

It seems I won't get to adding tests any time soon, as I'll be away from work for a couple of weeks. Do you prefer that I close this PR and come back with a PR that includes tests at some point in time or may I just leave it as is and add tests or another PR later?

splatch:
I would definitely welcome your contribution in 0.9 which is near by. It does not have a deadline, but several weeks might to long to wait.
Given that subscriptions were missing in 0.8 I don't mind getting it back without a very deep testing. This is a "preview of come back". I will leave PR open for a week and see if I will have time to bring necessary test before merging. If not then we will get it as-is.
One thing to note - did you sign Apache ICLA? If not please follow instructions on ASF page: https://www.apache.org/licenses/contributor-agreements.html

rmeister:
> One thing to note - did you sign Apache ICLA? If not please follow instructions on ASF page: https://www.apache.org/licenses/contributor-agreements.html

Ah shoot, I've looked for the ICLA on the PLC4X website, but obviously I haven't looked hard enough.
I have now signed and sent it to the secretary.

CCLA will follow in about two weeks.

splatch:
I confirm we received your contributor agreement. If you did this as part of your paid time as an employee then CCLA is welcome too.

Most likely I will run out of time this week to test your changes, so it will slip till weekend or next Monday. Nonetheless I keep an eye on it!

chrisdutz:
Hi @splatch ... did you manage to review this PR?
Was investing quite some time today to get dangling PRs back to develop

splatch:
@rmeister I've just did a test run of subscribe logic with TC 2 and it didn't fly.

Test code I made is following:
```
String fieldString = ".BYTE_DAT:BYTE";
System.out.println("Subscribe " + fieldString);
plcConnection.subscriptionRequestBuilder()
    .addChangeOfStateField("field1", fieldString)
    .build().execute().whenComplete((r, e) -> {
        if (e != null) {
            e.printStackTrace();
            return;
        }
        System.out.println("Subscribed");
        r.getSubscriptionHandle("field1").register(handler);
    });
```

Given symbol is declared in PLC program and it is resolved properly when I submit read request. Any ideas why it fails? I'll have another test with TC3.

sruehl:
for reference here is the implementation that was working before the mspec refactoring https://github.com/apache/plc4x/blob/release/0.6.0/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/connection/AdsTcpPlcConnection.java#L147

splatch:
@rmeister Hey, hope you have returned from your vacations. Can you confirm which TwinCat version you tested above logic with? At this moment I can't get this PR working with TC2.

rmeister:
@splatch Hey, well I wish it was vacation :sweat_smile:. I'll be away for at least two more weeks. I left the Beckhoff PLC at the office, so I can not repeat my test, but I can say it was TC 2(.11 probably).

Most probably there is a spot where I got confused over subscription name and ADS symbol name. If I remember correctly I chose both to be equal in my test.

rmeister:
Finally, I had some time to come back to this. Our TwinCat version is 2.11.2306 

@splatch I took your test code with this handler:
```
r.getSubscriptionHandle("field1").register(plcSubscriptionEvent ->{
	for (String fieldName : plcSubscriptionEvent.getFieldNames()) {
		System.out.println("String: " + plcSubscriptionEvent.getString(fieldName));
	}
} );
```
I didn't get any errors, so my theory that both arguments of `addChangeOfStateField` have to be the same is wrong. 
Did you get an error/exception or no response at all?

splatch:
> Finally, I had some time to come back to this. Our TwinCat version is 2.11.2306
> 
> @splatch I took your test code with this handler:
> 
> ```
> r.getSubscriptionHandle("field1").register(plcSubscriptionEvent ->{
> 	for (String fieldName : plcSubscriptionEvent.getFieldNames()) {
> 		System.out.println("String: " + plcSubscriptionEvent.getString(fieldName));
> 	}
> } );
> ```
> 
> I didn't get any errors, so my theory that both arguments of `addChangeOfStateField` have to be the same is wrong. Did you get an error/exception or no response at all?

As far I remember the 10 year old CX9000 I tested with did not return any answer nor error code. I have to assembly back my test board to re-issue tests. It could be that it is simply too old for that. ;-)

protocols/ads/src/main/resources/protocols/ads/ads.mspec
```diff
@@ -281,7 +281,8 @@
             // 4 bytes	The ADS server checks if the value changes in this time slice. The unit is 1ms
             [simple uint 32 'cycleTime']
             // 16bytes	Must be set to 0
-            [reserved   uint       128       '0x0000' ]
+            [reserved   uint       64       '0x0000' ]
```
sruehl, line 284: short question here: what was the reason for this change?

sruehl: other than that LGTM

chrisdutz: I think not all languages support 128bit integers

chrisdutz: So instead of one 128bit zero we now have two times 64bit

sruehl: golang and java do support it (BigInteger/big.Int)

rmeister: Good question. Reading my commit message again, the ParseException I got tells a different story: "Unsigned Big Integer can only contain max 64 bits"
I'll look if there is another way around that.

chrisdutz: I think I did a similar refactoring because I was running into problems ... so I wasn't randomly changing things ;-) In this case I thought fixing it wans't worth the effort. I mean .... it's for writing 16 zero bytes ... So I'm totally fine with that.

sruehl: ja let's use that then... I'll check later if we can support that (bit strange that it compiled before.)

PR #276, opened 2021-09-06
Change PLC4GO docs read syntax

NiklasMerz:
According to [this doc](https://github.com/apache/plc4x/blob/d767dc9b9160/plc4go/examples/read/hello_world_plc4go_read.go) the new syntax is `AddQuery`. 

`AddField` does not work for me after following the quick start


sruehl:
LTGM, thanks

sruehl:
as this is still a draft (which I didn't notice at first) I guess there will be more commits

sruehl:
github concurrency 😂

NiklasMerz:
I just marked it ready after my test was successful :-)

PR #288, opened 2021-12-01
Fix documentation for S7 short address format

markus-franke:
subtle difference making a big change ;-)


sruehl:
Thanks for the PR :)

PR #300, opened 2022-01-04
Bug/plc4x-330

hutcheb:
I haven't been able to replicate it, but have changed the port number for the subscription handler tests.


sruehl:
Is this done? Correct me if I'm wrong: I'm not sure if the last state was that you wanted to finish that.

hutcheb:
Looking at it, there's still a test disabled for OPCUA, which I believe is the cause for the issue in PLC4X-330

` public void multipleThreads()`

However this PR is a failed attempt to fix the jira issue, by changing the port number of the test opc ua server.
I'll delete this PR, but are you still having issues on your mac with this?

PR #305, opened 2022-01-19
Remove commons-beanutils

todoubaba:
Remove commons-beanutils dependency, because Android does not have class java.beans.PropertyDescriptor.


plc4j/spi/src/main/java/org/apache/plc4x/java/spi/configuration/ConfigurationFactory.java
```diff
@@ -178,63 +171,78 @@ private static Object toFieldValue(Field field, String valueString) {
             try {
                 ConfigurationParameterConverter<?> converter = converterClass.getDeclaredConstructor().newInstance();
                 if (converter.getType().isAssignableFrom(field.getType())) {
-                    return converter.convert(valueString);
+                    valueString = converter.convert(valueString).toString();
```
chrisdutz: IntelliJ is reporting that this value is never used. This is due to the fact that in line 179 you are always throwing an exception.

chrisdutz (changes requested):

chrisdutz:
Please have a look at the tweaks I did ... but thank you very much for your PR :-)

sruehl:
@todoubaba please have a look on my response on the mailing list

plc4j/spi/pom.xml
```diff
@@ -48,10 +48,6 @@
       <artifactId>netty-transport</artifactId>
     </dependency>
 
-    <dependency>
```
sruehl, line 51: please check if upstream can fix that.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/configuration/ConfigurationFactory.java
```diff
@@ -161,80 +150,107 @@ private static String getConfigurationName(Field field) {
...
+    private static void setFieldValue(Object instance, Field field, String valueString) throws IllegalAccessException {
         if (field == null) {
             throw new IllegalArgumentException("Field not defined");
         }
 
-        if (field.getAnnotation(ParameterConverter.class) != null) {
-            Class<? extends ConfigurationParameterConverter<?>> converterClass = field.getAnnotation(ParameterConverter.class).value();
+        try {
+            // Make the field accessible
+            field.setAccessible(true);
```
sruehl, line 166: if upstream disapproves we should at least use commons FieldUtils here

sruehl (changes requested):

chrisdutz:
> @todoubaba please have a look on my response on the mailing list

But you please also see what others have written ;-)

sruehl:
@todoubaba I removed the dependency in develop. It should no longer use beanutils

todoubaba:
@chrisdutz @sruehl Thank you all, it's a great project, and a great community.

chrisdutz:
Hi @todoubaba,
sorry for having wasted your time in preparing a PR ... I do think @sruehl 's solution should be the simpler one. I hope you don't mind that I simply close this PR without merging your changes. Please don't feel discouraged.

And of course we would be super excited if you could describe on the mailinglist, what you are doing with PLC4X. Cause most of us work on PLC4X but don't actually have direct use-cases, we just think that the world needs something like this project. Knowing what it's being used for is a huge motivation boost.

Thanks again,
Chris

todoubaba:
@chrisdutz It's OK. I am using PLC4X in a medical device，the host node is an industrial Android screen, and the slave nodes are single axis controllers. The protocol is CANOpen, and the transport is a customized CANTransport over ethernet.

splatch:
@todoubaba Does commons-lang work with Anadroid? If so then we still can amend this PR to rely on FieldUtils mentioned earlier.

todoubaba:
@splatch With my own patch, Plc4j works fine with Android. Not found any issues in commons-lang.

splatch:
@todoubaba Since you removed beanutils but still had commons-lang on classpath I believe we can get issue solved with `FieldUtils` so you will have "standard" release of PLC4X working under Android, WDYT?

chrisdutz:
@splatch ... this issue has been resolved as @sruehl already addressed and fixed the issue on the develop branch.

chrisdutz:
The problem was the usage of the BeanUtils and not the FieldUtils. So by switching on setting the fields in the configs via FIeldUtils (the annotation is field-based anyway) the code should be able to run on Android (As far as I understood it)

splatch:
Roger, I wasn't clear if we solved it for Lin or not.

todoubaba:
@splatch I updated the snapshot from https://repository.apache.org/content/repositories/snapshots/org/apache/plc4x/plc4j-spi/0.10.0-SNAPSHOT/plc4j-spi-0.10.0-20220124.174152-1.jar, it works fine with Android, the problem is solved. 

Thank you all.

chrisdutz:
Thanks for your feedback .. it's highly appreciated.

PR #317, opened 2022-02-18
Fix for failed opcua write attempts

splatch:
This commit relates to PLC4X-319, trying to cover faulty responses.


splatch:
Solved through #1108.

PR #318, opened 2022-02-18
fix(docs): Updated the pcap docs for when setting up a developer env

hutcheb:
The Npcap library seems to be maintained, however it looks like the license is a bit odd so we shouldn't use it.
This PR was to replace any comments about WinPcap with Npcap.


hutcheb:
The license for npcap cap is a bit odd, we probably shouldn't be using it.

PR #339, opened 2022-03-31
Fix casting to byte for integer greater than 127 (byte is signed numb…

alessandromnc94:
…er and for string greater than 127 characters it result to NegativeIndexException).

Edited only java version


ottobackwards:
Thanks for taking a look at this, can you please add a unit test for the behavior

chrisdutz (approved):
lgtm

PR #341, opened 2022-04-04
Feature/plc4py bh1

hutcheb:
Add Pluggy Framework
Fixed type hinting issues with the help of mypy


ottobackwards (commented):
This is really nice.

My only ask is that you doc string all the methods you have added, as it is inconsistently done here.

Also,  the 'hook namespace' in the drivers seems like it should be an interface ( or rather an abstract base class).  what do you think?  Can you do that with pluggy?

hutcheb:
Thanks Otto,

The Pluggy stuff definitely needs some work, I'm not happy with having to register the driver manually, kind of misses the point of using Pluggy.

I spent a bit of time trying to figure it out and moved on to have something that works. I'll take another look.

ottobackwards:
I don't think you have to register them manually, I think that they can register themselves using the setuptools / setup config.

So, if we start out with 1 plc4py whl file, it's setup can register all the distributed drivers like.

https://pluggy.readthedocs.io/en/stable/#loading-setuptools-entry-points

hutcheb:
Yeah found a way forward with adding the plugins to the poetry config.
I've also added an abtract class PlcConnectionLoader, almost feels a bit of a duplicate of the hookspec but I think it is better.

Yeah I've left off some of the docstrings, I've updated most but still left some of the stub files.

I'll rebase and merge it soon

ottobackwards:
sounds great

PR #348, opened 2022-04-11
Started with the PlcReadRequest

hutcheb:


hutcheb:
So I've ended up with abstract class such as this which allow you to pass return types, in this case it goes 3 or 4 deep before I create a concrete class. It seems to work but feels a bit clunky when I assigning multiple TypeVar's.

Just looking for opinions :)

`
T = TypeVar("T", bound=PlcResponse)


class PlcRequest(Generic[T], ABC):
    """
    Base type for all messages sent from the plc4x system to a connected plc.
    """

    @abstractmethod
    def execute(self) -> Future[T]:
        pass


U = TypeVar("U", bound=PlcFieldResponse)


class PlcFieldRequest(PlcRequest[U]):
    def __init__(self, fields: list[PlcField] = []):
        self.fields = fields

    @abstractmethod
    def execute(self) -> Future[U]:
        pass

    @property
    def field_names(self):
        return [field.name for field in self.fields]
`

hutcheb:
I also hate Circular dependancies, running into it with the PlcResponse and PlcRequest classes

hutcheb:
https://softwareengineering.stackexchange.com/questions/369146/how-to-avoid-bidirectional-class-and-module-dependencies

ottlukas:
> https://softwareengineering.stackexchange.com/questions/369146/how-to-avoid-bidirectional-class-and-module-dependencies

I liked accepted answer: 

- Hide it with interfaces
- Add an intermediary object

So are you going for Interfaces or a third object?

hutcheb:
I think an intermediary 'client' object makes more sense.
I've pulled out the PlcResponse and PlcRequest stuff from the opposite class.
In the Java version there's a PlcRequestContainer which seems to perform a similar function.

ottlukas (approved):

hutcheb:
Having trouble with pycharm and poetry for some reason, 
I'll merge this and swap to using a new build system.

ottlukas:
> Having trouble with pycharm and poetry for some reason, I'll merge this and swap to using a new build system.

Ok then lets bring this change to the mailing list - we already discussed that on slack and we should have that on the mailing list for confirmation. As soon as we have selected another build system

PR #351, opened 2022-04-19
Bug Fix. Open too many files after complete

Dustone-JavaWeb:
Shutdown the workergroup.  remove the final tag


JulianFeinauer (approved):
This change makes total sense and I wonder why this code wasnt in there at the first place. Many thanks @Dustone-JavaWeb for the contribution!

chrisdutz (approved):

PR #352, opened 2022-04-19
"Modbus plus"

splatch:
This pull request solves few errors I spotted while trying to use plc4j modbus stuff.
First of all, it adds support for String outputs from modbus requests, so `holding:3000:STRING[30]` will bring back up to 15 character text which wasn't covered properly by codegen till now. It works with `vstring` stuff Ben added a while ago.
Then there is second part which is actual modbus plus stuff. I provided sample request/responses I acquired from inverter, so its possible to see actual packets and payload mapping. For now I did stick with structs as identification answer is nested structure with variable list of elements.

### DIFF
--- a/protocols/modbus/src/main/resources/protocols/modbus/modbus.mspec
+++ b/protocols/modbus/src/main/resources/protocols/modbus/modbus.mspec
@@ -416,6 +416,15 @@
         ['WCHAR' List
             [array uint 16 value count 'numberOfValues']
         ]
+        ['STRING','1' STRING
+            [simple vstring '8' value encoding='"UTF-8"']
+        ]
+        ['STRING' List
+            [simple vstring '8 * numberOfValues' value encoding='"UTF-8"']
+        ]
+        ['WSTRING' List
+            [simple vstring '16 * numberOfValues' value encoding='"UTF-16"']
+        ]
     ]
 ]
 
--- a/code-generation/language-java/src/main/resources/templates/java/data-io-template.java.ftlh
+++ b/code-generation/language-java/src/main/resources/templates/java/data-io-template.java.ftlh
@@ -317,9 +317,14 @@ public class ${type.name} {
             return new PlcStruct(_map);
                     <#break>
                     <#case "List">
+                    <#if helper.getNonPrimitiveLanguageTypeNameForField(simpleField) == 'String'>
+            return new PlcSTRING(value);
+                    <#else>
             return new PlcList(value);
+                    </#if>
                     <#break>
                     <#default>
+                    // default ${case.name}
             return new Plc${case.name}(value);
                 </#switch>
             </#if>
--- a/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/ascii/protocol/ModbusAsciiProtocolLogic.java
+++ b/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/ascii/protocol/ModbusAsciiProtocolLogic.java
@@ -27,10 +27,9 @@ import org.apache.plc4x.java.api.model.PlcField;
 import org.apache.plc4x.java.api.types.PlcResponseCode;
 import org.apache.plc4x.java.api.value.PlcValue;
 import org.apache.plc4x.java.modbus.ascii.config.ModbusAsciiConfiguration;
-import org.apache.plc4x.java.modbus.base.field.ModbusField;
+import org.apache.plc4x.java.modbus.base.field.ModbusFieldBase;
 import org.apache.plc4x.java.modbus.base.protocol.ModbusProtocolLogic;
 import org.apache.plc4x.java.modbus.readwrite.*;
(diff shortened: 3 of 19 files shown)


splatch:
Seems C templates needs to be updated as well to handle string fields. Sadly I can't handle that part.

chrisdutz:
Ther test.mspec contains vstring fields ... could you please elaborate on what is actually not working?
In this case the test.mspec should be extended with an example that doesn't work. Then I'll take care of fixing this.

splatch:
> Ther test.mspec contains vstring fields ... could you please elaborate on what is actually not working? In this case the test.mspec should be extended with an example that doesn't work. Then I'll take care of fixing this.

I think it might be related to use of vstring inside data-io:
```
        ['STRING' List
            [simple vstring '8 * numberOfValues' value encoding='"UTF-8"']
        ]
        ['WSTRING' List
            [simple vstring '16 * numberOfValues' value encoding='"UTF-16"']
        ]
```

For Java problem was that `case` statement was catching this as `List`, so there is a need to verify plain type read from buffer. If its `String` then, output is not an array but, just an string.

chrisdutz:
Your mspec fragment looks a bit fishy in general ... the type you define as "List" and not as "String" and where does the "numberOfValues" come from?

splatch:
> Your mspec fragment looks a bit fishy in general ... the type you define as "List" and not as "String" and where does the "numberOfValues" come from?

Its being called when you declare field data type as `STRING[10]`.

chrisdutz:
"List" would require a field of type array labeled "value" ....  So I think this should be "STRING" instead ... but the variable of the length I can't see where it's coming from.

splatch:
I get your point as it is tricky. For me with high level language it makes sense to avoid construction of a `String[] {"a", "b", "c", "d" }` and return directly a String "abcd" to the caller.

protocols/modbus/src/main/resources/protocols/modbus/modbus.mspec
```diff
@@ -416,6 +416,15 @@
         ['WCHAR' List
             [array uint 16 value count 'numberOfValues']
         ]
+        ['STRING','1' STRING
```
chrisdutz, line 419: Ok ... so in this case the numberOfValues is 1 ... and your string consists of only one utf-8 char ... for this case I would use "string 8" instead of this. But I think you need to somehow know how long the string should be ... the numberOfValues shouldn't be the lenght of the string, but the number of strings ... usually I have one internal field that contains the string lenght and then I use that in the vstring.

protocols/modbus/src/main/resources/protocols/modbus/modbus.mspec
```diff
@@ -416,6 +416,15 @@
...
+            [simple vstring '8' value encoding='"UTF-8"']
+        ]
+        ['STRING' List
+            [simple vstring '8 * numberOfValues' value encoding='"UTF-8"']
```
chrisdutz, line 423: Here you are creating only one single element field of type string that you want to put in a list ... I think you should instead split this up into two cases: numberOfValues = 1, where the type is STRING and one where it's not == 1 where the type is List and the value field then needs to be an array.

chrisdutz (changes requested):

protocols/modbus/src/main/resources/protocols/modbus/modbus.mspec
splatch: It doesn't make much sense, cause output should be a `String` anyway. What's the point of returning a text in chunks to the caller if it can be merged and read at once beforehand?

chrisdutz: I think you are mis-understanding the concept here ... 
A field of type "string" or "vstring" can contain a string of multiple characters. But with "STRING[10]" we are not reading one String with a length of 10 characters, but 10 strings of individual length (In S7 this would be 255 chars long) ... if you want to explicitly control the string lenght that you want to read use round bracets ... so STRING(10)[20] would read 20 strings of each 10 chars length. You are currently mixing up the array size and the lenght.

splatch: Ok, it makes more sense then, will check out how rounded brackets go and update/revert template change if needed.

chrisdutz: Have a look at the S7 protocol ... there we're already using the string-length.

splatch: Looking at S7 fields I see that its a special case regexp which is not going to fly with modbus unless we modify its definition. Will try cooking something there.

chrisdutz:
@splatch What's the state of this?

splatch:
Hey Chris,
It is stale. Protocol wise, it was rather a minor change.  I have no more access to inverter which provided answers, so all I have are samples (device identification request & response) attached to this PR.

hutcheb:
Closing due to inactivity

PR #354, opened 2022-04-21
Fix/plc4py_pipenv

hutcheb:


hutcheb:
Using pipenv to build is too slow and adds an unnecessary dependency.

PR #363, opened 2022-05-12
fix(plc4go): fix the error of value out of range when converting an i…

hongjinlin:
…n-range number string to IEC61131 unsigned integer or WORD-related type.


sruehl:
thx for the PR

PR #384, opened 2022-06-15
Updated java CIP write dataSize to match read dataSize

AndyGrebe:


chrisdutz:
Could you please add a bit more information in this PR (here in Github) to what the PR addresses?

hutcheb:
It looks like it is trying to handle fields a little better by taking into account members of structures.

I had done some work with a logix controller on the logix_develop branch. Came to the conclusion that it isn't massively different to what gets used in the eip driver. But we'd probably need to have some sort of device identification to account for endianess and what each device supports.

I went through and updated the mspec to have more information but I think it is still compatible with eip mspec (but alot of the logic would need to change).

Within https://github.com/apache/plc4x/blob/logix_develop/plc4j/drivers/logix/src/main/java/org/apache/plc4x/java/logix/readwrite/protocol/LogixProtocolLogic.java

I've updated the toAnsi method to account for structures, but it uses a lot of the new structures within the updated mspec.

chrisdutz:
So is this in a mergable state? I think you're probably the best to say anything about this. If it is ... please be so kind an merge it.

AndyGrebe:
Hi,

Sorry for the late reply.  The path size for writing the structure members
was the wrong size.  It was fixed on the read side, but not the write.

Thanks,
Andy

hutcheb:
Hi Andy,

Yeah the PR looks good., thank you. It looks like this is your first contribution to not only the ASF but also open source on GitHub, Congratulations.

As part of the ASF requirements though, a license agreement (ICLA) will need to be signed and sent to the ASF. All the information you'll need is at [license agreement](https://www.apache.org/licenses/contributor-agreements.html#clas).

Some additional information for new comers can be found [here](https://community.apache.org/)

If you have any questions at all let us know, looking forward to working with you.

Ben

chrisdutz:
I think the contribution is small enough to be ok to merge without ... however if you plan on actively contributing more (which I really hope), then it would be cool, if you could file an ICLA as Ben pointed out

AndyGrebe:
I'll make sure to fill the ICLA.  I have a feeling we will be working
together quite a bit.

PR #386, opened 2022-06-18
Fix bug with select returning prematurely when device not sending any data

vmpn:
Selector API: https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Selector.html for select() method states:

    `returns only after at least one channel is selected, this selector's wakeup method is invoked, or the current thread is interrupted, whichever comes first.`

for the select(long timeout) it states:

    `timeout - If positive, block for up to timeout milliseconds, more or less, while waiting for a channel to become ready; if zero, block indefinitely; must not be negative`

Current implementation causes premature return from select when serial port is not receiving any data and wakeup() method does impact select() wait behavior as expected by the API

This PR changes the behavior of sleep(), sleep(long timeout) and wakeup() of SerialPollingSelector to be compliant with the APIs, removes unnecessary wake-ups  and addresses warnings issued by netty.

SelectorTest unit test can be used to demonstrate the issue with existing code and lack of the same after changes to the SerialPollingSelector class


sruehl:
thanks for the PR @vmpn

sruehl:
once github becomes usable again I'll merge it...

PR #405, opened 2022-06-26
feat(eip): update eip driver to include an option for connected messaging.

hutcheb:
Update the EIP driver to include options for Little/Big endian and using connected messaging.


hutcheb:
@AndyGrebe It'd be great if you could take a look at this PR.

A new driver has been added for logix controllers. The connection string format will be something like 
`logix://192.168.1.1`

This should use Little Endian and the Field format now no longer requires a leading %.

chrisdutz:
@hutcheb what do you think? Is this PR obsolete?

PR #416, opened 2022-06-28
fixed copy-paste error in s7 config options

tvwy:
There has been a copy-paste bug in s7 options factory. "local-rack" option has been used for rack & slot variable. same with "remote-rack". 

I'm not a big go developer, auto formatter messed up formatting in import statement...


sruehl:
Indeed good catch, thanks for the PR 👍🏼

PR #257, opened 2021-06-24
Feature/nifi record

inigoao:
Nifi feature, record oriented processor
This PR substitutes former one (https://github.com/apache/plc4x/pull/242) to make clearer the code review.

PR #439, opened 2022-08-02
Apache Nifi plc4x record processor 2

inigoao:
This PR contains the updated code for Nifi Record oriented processor, substituting the former one (#257).


inigoao:
Hi all!

As we discussed on the dev-mail-list, we removed the dynamic address functionality by leaving PlcConnectionString and PlcAddressString properties static (hardcoded on the processor, not updated by incoming flowfile). 

Also we added a new method "parseAddressString()" on org.apache.plc4x.nifi.util.Plc4xCommon class that should ideally be reviewed. This method allows to extend PlcAddressString property format on special cases where the protocol syntax could create a conflict with the parsing the processor does. 
The syntax the processor expects for this property is: varName1=varAddress1;varName2=varAddress2
Using OPCUA driver we found this conflict, where a variable had the following name and address format: MaxCurrentImax=ns=2;i=33
For allowing these characters on the varAddress we used the following approach: 'MaxCurrentImax'='ns=2;i=33'
So the idea behind the "parseAddressString()" method is to detect wether the protocol is OPCUA (or any other that could be included), to force the following variable address syntax: 'varName1'='varAddress1';'varName2'='varAddress2'

This is just a proposal, maybe there are better ways to approach this. 
Let us know your opinion!

ottobackwards:
Can you post a discussion to the dev@plc4x.apache.org mailing list about the address format problem?  I would like to get input from everyone.

ottobackwards:
Please rename this pr to something like "Apache Nifi plc4x record processor 2"

plc4j/integrations/apache-nifi/README.md
```diff
@@ -0,0 +1,90 @@
...
+- *Schema Name:* ${schema.name}
+- *Schema Version:* No value set
+- *Schema Branch:* No value set
+- *Schema Text:* ${avro.schema}
+- *Compression Format:* NONE
+- *Cache Size:* 1000
+- *Encoder Pool Size:* 32
+
+
+The output flowfile will contain the PLC read values. This information is included in the flowfile content, following the Record Oriented presentation using an **schema** and the configuration specified in the Record Writer (format, schema inclusion, etc). In the schema, one field will be included for each of the variables defined in the Processor's  *PLC resource address String:* property, taking into account the specified datatype. Also, a *ts* (timestamp) field is additionally included containing the read date. An example of the content of a flowfile for the previously defined properties:
```
ottobackwards: a schema

plc4j/integrations/apache-nifi/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+[ {
```
ottobackwards, line 56: I'm confused, the variables above don't have the "_type" on them.  Is that added? or are these names wrong?  They should match

plc4j/integrations/apache-nifi/nifi-plc4x-nar/pom.xml
```diff
@@ -18,146 +18,188 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.apache.plc4x</groupId>
+			<artifactId>plc4j-driver-knxnetip</artifactId>
+			<version>${nifi-plc4x.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.plc4x</groupId>
+			<artifactId>plc4j-driver-modbus</artifactId>
+			<version>${nifi-plc4x.version}</version>
```
ottobackwards, line 127: remove commented entries

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
```diff
@@ -0,0 +1,128 @@
...
+  	// inigo private AtomicReference<Map<String, RecordSchema>> map 
+
+    public Plc4xReadResponseRecordSet(final PlcReadResponse readResponse) throws IOException {
+        this.readResponse = readResponse;
+        moreRows = true;
+        
+        logger.debug("Creating record schema from PlcReadResponse");
+        Map<String, ? extends PlcValue> responseDataStructure = readResponse.getAsPlcValue().getStruct();
+        rsColumnNames = responseDataStructure.keySet();
+
```
ottobackwards: english please

ottobackwards (commented):
A couple of comments.
Also, instead of a single property with multiple addresses, why don't you support dynamic properties, where the user could just add multiple instances?  Each address would be it's own  and you wouldn't have to coerce

https://nifi.apache.org/docs/nifi-docs/html/developer-guide.html#dynamic-processor-properties

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
chrisdutz: Which word in this is not english?

plc4j/integrations/apache-nifi/README.md
inigoao: done, thanks

plc4j/integrations/apache-nifi/README.md
inigoao: I think you refer to the table at the end, where all the avro dataypes have a "_type" suffix on it, is that right? I corrected it to leave only datatype names

plc4j/integrations/apache-nifi/nifi-plc4x-nar/pom.xml
inigoao: done, thank you

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
inigoao: yes I forgot some comments on spanish I used while developing... done, thanks

ottobackwards:
can you have a look at https://github.com/apache/plc4x/pull/439/checks?check_run_id=7654047052, and where they are direct dependencies you have added try updating the versions to the suggested versions?

inigoao:
hi all, 
we took a look to the dynamic properties, now the processor supports both methods to specify variable addresses: using PlcAddressString property and using dynamic properties. We added another property (selector), to select the desired method to specify these variables. In this way, we can test both methods, and in the future we can decide which one we prefer

also, we rewrote the Plc4xCommon.parseAddressString() method, using StringUtils library. Here, another approach was to parse the string as a CSV to get the objects. With the later, we did some tests, but didn't added it to the commit, maybe you can take a look and give some feedback here.

lastly, we looked at the dependency version change suggestions ( https://github.com/apache/plc4x/pull/439/checks?check_run_id=7654047052 ), we found that some nifi dependencies are using avro 1.8.1 which is causing the conflicts, although it is supposed to be fixed already..  we maintained it, and added a exclusion for the commons-compress dep, which I think know is provided by the parent projects.

let us know what you think, thank you!

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/util/Plc4xCommon.java
```diff
@@ -258,34 +259,43 @@ else if (value.isTime())
 	 * @param addresses
 	 * @return
 	 */
-	public static Map<String, String> parseAddressString(String connectionString, PropertyValue addresses){
+	public static Map<String, String> parseAddressString(String connectionString, PropertyValue addressesProp){
 		Map<String, String> addressMap = new HashMap<>();
-		//TODO OPCUA var adresses sintax
+		String addresses = addressesProp.getValue();
+
```
ottobackwards: This is a big yellow warning sign/code smell however you want to call it.
The insistence on having an option to do something in a way that is more complicated, error prone, and requires this kind of stuff to work is an anti-pattern.

Also, it seems you want to add *more* properties, to help control how this complicated code works, to make it even more complicated.

I will not ask you to change it however.

ottobackwards:
@chrisdutz I'll leave final review to you

ottobackwards:
I'm +1 so we can get this in and going.  If this existed, and this pr was adding the address string, I would be -1 however.

ottobackwards:
@inigoao please excuse the terseness of my review comments.  Let me say thank you for the contribution, and you all did a great job!

inigoao:
we have updated the code to remove the Plc Address String prop, keeping only the Dynamic Properties approach, let us know what you think

PR #494, opened 2022-09-09
fix: opm collection field set value error

fangzhengjin:
example:
```java
@PlcField("holding-register:1[4]")
private List<Integer> test4;
```


sruehl:
thanks for the PR. Can you (@fangzhengjin) please add a unit-test case for that?

chrisdutz (approved):

chrisdutz:
I'll re-write the change after the tag/tagAddress refactoring ...

chrisdutz:
Re-implemented the change

PR #558, opened 2022-10-24
Feature: Add connection option to connect with CPs

JulianFeinauer:


splatch:
@JulianFeinauer Does it apply to later releases? If so, can we ensure its available there too?

chrisdutz:
I would assume the same should apply also for the newer versions ... we should double-check

chrisdutz:
We don't have a "device-group" setting in the newer versions ... and probably, if we added it, we should add variants for local and remote "device-group" ... "local-device-group" defaulting to PG_OR_PC and "remote-device-group" defaulting to OTHERS.

JulianFeinauer:
Indeed, good point. We came across this when connecting to an old s7 300 with a Communication processor. I don't know the theory well enough but it seems in this case we do not need rack but slot and device group setting

Issue #568, opened 2022-11-02
(plc4j opcua driver) Unable to read when node id contains unicode characters


modraedlau:
Unable to read when node id contains unicode characters. 
For example: 
The nodeid is `ns=2;s=设备1`, the read result will be `NOT_FOUND`.

Look at the generated class `org.apache.plc4x.java.opcua.readwrite.PascalString`, the `getStringLength` equals to string length, should this be equal to `getBytes().length`?

hutcheb:
Thanks modraediau, You are right, the OPCUA standard doesn't seem to define the format of the Node Id string Identifier.

It would be great if you could create a PR for this. The PascalString class is defined in the mspec/xslt file here https://github.com/apache/plc4x/blob/develop/protocols/opcua/src/main/xslt/opc-manual.xsl

If you have any issues, or need some help let us know.

If you aren't in a position to create a PR, I'd be more than happy to test the change in the next few days.

splatch:
@modraedlau It might be related to utf-8 vs utf-16. What's your platform jvm setting in this regard?

modraedlau:
I submitted a pull request：[fix(plc4j/opcua): not to get String.length but to calculate length in...](https://github.com/apache/plc4x/pull/668)

PR #668, opened 2022-11-23
fix(plc4j/opcua): not to get String.length but to calculate length in…

modraedlau:
Not to get String.length but to calculate length in UTF-8 for Pascal string length (support node id contains unicode characters).

fix the issue [https://github.com/apache/plc4x/issues/568](https://github.com/apache/plc4x/issues/568)

Thanks for reviewing the code.


hutcheb:
Thanks modraedlau, It seems to work as expected.

On a side note emoji's are now also supported as well
image

Issue #710, opened 2022-12-22
modbus-tcp Write Request coil always set to false


SteinOv:
This is a bug that is new in release 0.10.0 and did not exist in 0.9.1.

When doing a write request for a modbus coil, the coil is always set to false, no matter the value given.

it can be reproduced by running the plc4j-hello-world-plc4x-write example with the following arguments:
`--connection-string "modbus-tcp:tcp://localhost:502" --field-addresses coil:100 coil:101 --field-values true true`
Even though both values are set to true, the actual coils will be set to false. At first I tried it with code I wrote myself, where I encountered the issue, so the problem is not in the example code.

It occurred both when using an actual modbus slave and another library that emulates a slave:
```
import java.util.Arrays;

import com.ghgande.j2mod.modbus.procimg.ProcessImage;
import com.ghgande.j2mod.modbus.procimg.SimpleDigitalOut;
import com.ghgande.j2mod.modbus.procimg.SimpleProcessImage;
import com.ghgande.j2mod.modbus.slave.ModbusSlave;
import com.ghgande.j2mod.modbus.slave.ModbusSlaveFactory;

import lombok.extern.slf4j.Slf4j;

@Slf4j
public class Slave {
    private static final boolean[] state = new boolean[1000];

    public static void main(String[] args) {
        Arrays.fill(state, true);
        
        ModbusSlave slave = null;
        try {
            ProcessImage image = createProcessImage();
            slave = ModbusSlaveFactory.createTCPSlave(502, 5);
            slave.addProcessImage(1, image);

            // Start the slave listening on the port
            slave.open();
            while (true) {
                try {
                    Thread.sleep(100);
                    for (int i = 0; i < 1000; i++) {
                        // Log if coil changed
                        if (image.getDigitalOut(i).isSet() != state[i]) {
                            state[i] = !state[i];
                            log.info("\n===\naddr {} changed to: [{}]\n===\n", i, state[i]);
                        }

                    }
                } catch (InterruptedException e) {
                    Thread.currentThread().interrupt();
                }
            }

        } catch (Exception e) {
            log.error("Modbus Slave error", e);
        } finally {
            if (slave != null) {
                // Close the slave
                slave.close();
            }
        }
    }

    private static ProcessImage createProcessImage() {
        SimpleProcessImage image = new SimpleProcessImage();
        // Create coils 0-999
        for (int i = 0; i < 1000; i++) {
            image.addDigitalOut(i, new SimpleDigitalOut(state[i]));
        }
[...]

PR #711, opened 2022-12-23
fix(protocols/modbus): fix write requests for coils always set to false (#710)

SteinOv:
see [issue-710](https://github.com/apache/plc4x/issues/710)

### DIFF
--- a/plc4j/drivers/modbus/src/main/generated/org/apache/plc4x/java/modbus/readwrite/DataItem.java
+++ b/plc4j/drivers/modbus/src/main/generated/org/apache/plc4x/java/modbus/readwrite/DataItem.java
@@ -48,11 +48,14 @@ public class DataItem {
 
       // Reserved Field (Compartmentalized so the "reserved" variable can't leak)
       {
-        short reserved = /*TODO: migrate me*/ /*TODO: migrate me*/
-            readBuffer.readUnsignedShort("", 7);
-        if (reserved != (short) 0x00) {
+        int reserved = /*TODO: migrate me*/ /*TODO: migrate me*/ readBuffer.readUnsignedInt("", 15);
+        if (reserved != (int) 0x0000) {
           LOGGER.info(
-              "Expected constant value " + 0x00 + " but got " + reserved + " for reserved field.");
+              "Expected constant value "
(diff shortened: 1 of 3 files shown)


chrisdutz:
I think in this case it would then work for the coils, but break for the registers, as these are 16 bit units.

I think a better solution would be to have either a dedicated DataIo for coil data (However this would have very few cases) or to not use the DataIo component for parsing the coil data, but to implement it manually (which shouldn't be much work) for this case (One could argue, that reading 8 coils and having that interpreted as a signed 8-bit integer would no longer be supported by that, but that's probably something I would be able to live with)

chrisdutz (changes requested):
This change will definitely break boolean support for registers.

SteinOv:
What kind of solution did you have in mind and why would it break the functionality you're describing?
I thought maybe adding an if statement to the fromPlcValue() method in ModbusProtocolLogic: `if (tag instanceof ModbusTagCoil && plcValue instanceof PlcBOOL)` and then manually parsing the value there.
What do you think?

chrisdutz:
Oh ... thought I had mentioned this on the main PR comment:
Reason is that for Registers it sort of expects 16 bit ... if we reduce this to 8 again, we will only be able to detect the numeric value of 8 as a true.
I think the ideal solution would be to have:
- A separate data io for coils
- Extend the existing data io with an additional property, if this is a coil or a register and to have various options for dealing with it
- Simply manually implement the parsing of coils

My prefered option would be the 3rd ... manually implementing the parsing of coils.

SteinOv:
Sorry, I wasn't clear. My comment was a response to your first comment. You said that manually implementing the parsing the coil data would break being able to interpret 8 coils as an signed 8-bit integer. I was wondering what kind of solution for manually implementing the parsing of the coil data you had in mind and how this would break that functionality.
I also thought about it myself and suggested a way to manually implement the parsing, see my previous comment.

SteinOv:
See my latest commit for what I had in mind.

chrisdutz:
Hi,

Yeah ... this should be a good solution. And it's way less complicated ... but happy you at least could have a look at the general concept of our code-generation ;-)

Chris

PR #721, opened 2022-12-30
Limit modbus quantity

nielsbasjes:
When requesting too many registers in a single modbus request the request will fail with a very unclear modbus error from the actual device at hand.

The modbus specification ([pdf](https://www.modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf)) documents limits of at most 125 registers or 2000 coils in a single request which matches with the limits I have observed in the modbus enabled appliance I have here.

This change will simply let the application fail before making the connection instead of the actual device returning a very unclear modbus error.


chrisdutz:
Hi Niels,

thanks for that ... I guess this would be a good first step, but I think the ideal solution would be to allow using more registers and coils in Modbus using the PLC4X API and to internally split them up into multiple requests. I think if you have a look at the S7Optimizer, this should be the best source of how to achieve this, however I guess as we're not taking about calculating packet sizes of the response it should be A LOT simpler. 

Would you be interested in taking on this feature?

Chris

chrisdutz (approved):

nielsbasjes:
I was also thinking about such an option.
Because I'm not familiar with the code yet I opted for this simpler fix as a first step.
I'll have a look at the class you mentioned and see if I can make this happen. 
No promises yet.

PR #724, opened 2023-01-02
fix: Improve java example

nielsbasjes:
Some improvements in the example:
- No need for close in a try-with-resources block.
- Move the exit outside the try-with-resources block. 
- Use the CompletableFuture (which is the actual return type of execute) which has a `get()` that waits for the completion --> no need for an unreliable sleep.
- Document the need for the exit explicitly


chrisdutz:
Well admittedly the sleep doesn't make any sense with the async stuff commented out. 
Not quite sure why it's commented out at all. If we call "get" then the whole point of doing it asynchronously is lost.

But I think it makes sense to add a second example demonstrating the async behaviour. Perhaps with a hook that waits on a signal (strg+c) to shut down.

chrisdutz:
And expanding all of the code ... yes ... your soltion is nicer ... just noticed it was a new future you were getting ;-)
So yes ... I think that option is nicer.

nielsbasjes:
@chrisdutz @ottlukas I just noticed this is still open. 
Do you want to use this? Or are changes needed from my end? Or is this to be closed without being merged? Or ...

chrisdutz:
Sorry ... a bit turbulent on my side ... just merged it.

PR #732, opened 2023-01-08
fix(plc4j/modbus): Cleanup of ModbusTag

nielsbasjes:
This is my proposed set of (what I think are) improvements for the ModbusTag code.
Summary:
- The equals and hashcode were incorrect because the name of the actual class also matters.
- My take on reducing the confusion around the -1 offset and the code complexity around all of this:
  - Simplified the code for readability
  - There is now a getLogicalAddress which returns the address the user configured.
  - The getAddressString returns a string that parses (which was not the case) AND yields an identical new tag when parsed (which was not the case: was shifted by 1 most of the time).
  - A more extensive set of tests that verifies all of this and ensures all supported formats for all tags yield the correct tag that is identical regardless of the used format.

Looking forward to your feedback.

### DIFF
--- a/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/base/tag/ModbusTag.java
+++ b/plc4j/drivers/modbus/src/main/java/org/apache/plc4x/java/modbus/base/tag/ModbusTag.java
@@ -48,8 +48,6 @@ public abstract class ModbusTag implements PlcTag, Serializable {
 
     private final ModbusDataType dataType;
 
-    private final String addressStringPrefix;
-
     public static ModbusTag of(String addressString) {
         if (ModbusTagCoil.matches(addressString)) {
             return ModbusTagCoil.of(addressString);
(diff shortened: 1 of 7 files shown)


chrisdutz (approved):
Instead of using a private field "addressStringPrefix", I would probably have added an abstract protected method "getPrefix()" which each sub-type needs to implement and return. But this is just a tiny nitpick as it simply saves a few byte of ram for every instance. 
But the rest really looked good to me :-)

chrisdutz:
And just asking ... would you feel able to do the same for the PLC4Go and/or PLC4C implementation? We're trying to keep them as in-sync as possible ... if not, that's no issue ... just thought I'd ask ;-)

nielsbasjes:
I have put up some additional changes with a getAddressStringPrefix() method.

About the C and Go code ... I have never written any Go code yet and the last time I touched C/C++ was > 20 years ago. At this point I don't even know the basics of the build systems that are used (I assume things have changed since 2003).

chrisdutz:
Thank you for those changes :-) 

And I guess ... if you would like to get started in Plc4go, I would be happy to assist you ... PLC4C right now is probably not really worth the effort. It's a pretty "experimental" thing which might even be replaced by PLC4Rust once that's done.
If you're not interested, no worries :-)

nielsbasjes:
Usually I'm happy to contribute. 
Right now I would like to focus on the actual application code I'm building (in Java).
So currently I won't be digging in to the Go/C/Rust code.

PR #747, opened 2023-01-17
New Implementation of the Connection-Cache

chrisdutz:
New-implementation of the Connection Cache and removal of both the old connection-cache as well as the connection-pool.

In addition some types in the API module were updated:

- The PlcDriverManager was renamed to DefaultPlcDriverManager
- A new Type PlcDriverManager
- A new Type PlcConnectionManager
- PlcDriverManager contains a getConnectionManager() method
- PlcDriverManager contains a static method getDefault() which returns an instance of the DefaultDriverManager
- DefaultPlcDriverManager implements both interfaces


plc4j/api/src/main/java/org/apache/plc4x/java/DefaultPlcDriverManager.java
```diff
@@ -32,19 +34,19 @@
 import java.util.ServiceLoader;
 import java.util.Set;
 
-public class PlcDriverManager {
+public class DefaultPlcDriverManager implements PlcDriverManager, PlcConnectionManager {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(PlcDriverManager.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(org.apache.plc4x.java.DefaultPlcDriverManager.class);
```
sruehl:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPlcDriverManager.class);
```

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcConnectionManager.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Connects to a PLC using the given plc connection string using given authentication credentials.
+     *
+     * @param url            plc connection string.
+     * @param authentication authentication credentials.
+     * @return PlcConnection object.
+     * @throws PlcConnectionException an exception if the connection attempt failed.
+     */
+    PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException;
```
sruehl, line 43: @chrisdutz we should some day think about using the option pattern from Golang here to (the ones with the varargs options). This way we could combine the getConnection into one method.

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriverManager.java
```diff
@@ -16,17 +16,25 @@
...
+        return new DefaultPlcDriverManager();
+    }
+
+    Set<String> listDrivers();
+
+    PlcDriver getDriver(String protocolCode) throws PlcConnectionException;
+
+    PlcDriver getDriverForUrl(String url) throws PlcConnectionException;
 
-    PlcConnection create() throws PlcConnectionException;
```
sruehl, line 30: I wonder if we should leave the create method as deprecated default implementation:
```
default PlcConnection create() {
    return getConnectionManager().create();
}
```

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
```diff
@@ -136,11 +136,11 @@ public void onConnect(ConversationContext<AmsTCPPacket> context) {
         // otherwise just mark the connection as completed instantly.
         setupAmsRouteFuture.whenComplete((unused, throwable) -> {
             if (!configuration.isLoadSymbolAndDataTypeTables()) {
-                future.completeExceptionally(new PlcConnectionException(
-                    "Lazy loading is generally planned, but not implemented yet. " +
-                        "If you are in need for this feature, please reach out to the community."));
-            }
-            //if (configuration.isLoadSymbolAndDataTypeTables()) {
+                context.fireConnected();
+//                future.completeExceptionally(new PlcConnectionException(
```
sruehl, line 143: is the outcommmented code here an accident?

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcConnectionManager.java
chrisdutz: Well that is incompatible with the try-with-resources pattern ...

chrisdutz: At least I think so

plc4j/examples/hello-ads-telemetry/pom.xml
```diff
@@ -40,13 +43,31 @@
...
+        <configuration>
+          <usedDependencies combine.children="append">
+            <usedDependency>org.slf4j:log4j-over-slf4j</usedDependency>
+          </usedDependencies>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
```
sruehl, line 73:
```suggestion
</project>

```

plc4j/examples/hello-ads-telemetry/src/main/resources/logback.xml
```diff
@@ -27,7 +27,7 @@
     </encoder>
   </appender>
 
-  <root level="error">
+  <root level="info">
```
sruehl, line 30:
```suggestion
  <root level="error">
```

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
```diff
@@ -0,0 +1,119 @@
...
+    public CachedPlcConnectionManager(PlcConnectionManager connectionManager, Duration maxLeaseTime, Duration maxWaitTime) {
+        this.connectionManager = connectionManager;
+        this.maxLeaseTime = maxLeaseTime;
+        this.maxWaitTime = maxWaitTime;
+        this.connectionContainers = new HashMap<>();
+    }
+
+    public PlcConnection getConnection(String url) throws PlcConnectionException {
+        ConnectionContainer connectionContainer;
+        synchronized (connectionContainers) {
```
sruehl, line 63: an alternative here might be the use of the ConcurrentHashMap

plc4j/tools/connection-cache/src/test/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * This is the simplest possible test. Here the ConnectionManager is used exactly once.
+     * So not really much of the caching we can test, but it tests if we're creating connections the right way.
+     *
+     * @throws PlcConnectionException something went wrong
+     */
+    @Test
+    public void testSingleConnectionRequestTest() throws PlcConnectionException {
+        PlcConnectionManager mockConnectionManager = Mockito.mock(PlcConnectionManager.class);
```
sruehl, line 41: There is also the Mockito Annotation which is pretty cool to use

plc4j/tools/connection-cache/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,9 @@
     </encoder>
   </appender>
 
-  <root level="error">
+  <logger name="org.apache.plc4x.java.utils.cache" level="debug"/>
+
+  <root level="info">
```
sruehl, line 32:
```suggestion
  <root level="error">
```

plc4j/tools/connection-cache/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,9 @@
     </encoder>
   </appender>
 
-  <root level="error">
+  <logger name="org.apache.plc4x.java.utils.cache" level="debug"/>
```
sruehl, line 30:
```suggestion
  <logger name="org.apache.plc4x.java.utils.cache" level="error"/>
```

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
chrisdutz: No ... it's a first step to make it possible to still connect to an ADS without loading the symbol table (if you just use raw addresses) ...

plc4j/tools/scraper-ng/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.plc4x</groupId>
+      <artifactId>plc4j-driver-mock</artifactId>
+      <version>0.11.0-SNAPSHOT</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+</project>
```
sruehl, line 110:
```suggestion
</project>

```

plc4j/tools/scraper/pom.xml
```diff
@@ -100,12 +96,6 @@
...
     </dependency>
-    <dependency>
-      <groupId>org.apache.plc4x</groupId>
-      <artifactId>plc4j-connection-cache</artifactId>
-      <version>0.11.0-SNAPSHOT</version>
-      <scope>compile</scope>
-    </dependency>
   </dependencies>
 
 </project>
```
sruehl, line 101:
```suggestion
</project>

```

plc4j/tools/scraper-ng/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    <version>0.11.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>plc4j-scraper-ng</artifactId>
+
+  <name>PLC4J: Tools: Scraper (NG)</name>
+  <description>Utility to efficiently collect a large number of items on multiple devices by different triggers.</description>
+
+  <dependencies>
+    <!--Jackson-->
```
sruehl, line 36: maybe remove those outcommented deps

sruehl (approved):
Left some suggestions which should be easy to apply

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcConnectionManager.java
sruehl: Don't think that should be a Problem but maybe I don't see it :D

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriverManager.java
chrisdutz: What for? I mean ... we're obviously not using it.

plc4j/examples/hello-ads-telemetry/pom.xml
chrisdutz: Seriously? Not willing to do a discussion on the usage of new-lines.

plc4j/examples/hello-ads-telemetry/src/main/resources/logback.xml
chrisdutz: I'm using the logging to output information on INFO level ... what good would the application be, if we set the log level to Error and just stripped all usable output? If yes, then you should have suggested a logger entry to explicitly set the log level of the application to info, but this change doesn't make any sense to me.

plc4j/examples/hello-ads-telemetry/pom.xml
sruehl: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriverManager.java
sruehl: because it is part of the API package an might break exiting deployments from customers.

plc4j/examples/hello-ads-telemetry/src/main/resources/logback.xml
sruehl: Couple of months back I turned them all up to error as we just exceeded the logoutput because of the size of our build. Those changes where all in the test-logback.xml. Just noticed that this is the main logback file so that should be fine so nvm

plc4j/api/src/main/java/org/apache/plc4x/java/api/PlcDriverManager.java
chrisdutz: Let's agree on not calling them "customers" as they are missing one important aspect of a typical customer relation.
I would say that "customers" have the chance to come forward and announce this would break things for them. I would consider it important to have backward compatibility once we are at 1.0.0. But I wouldn't object your proposal. Feel free to add it.

PR #752, opened 2023-01-18
Feature/nifi integration type mapping

QuanticPony:
Improvements in type mapping from Plc4x to Avro. Added unit test for record source processor.

Changes in mapping:
* PlcBYTE:  string -> bytes
* PlcLINT:    string -> long
* PlcREAL:  double -> float
* PlcLREAL: string -> double

General improvements in `org.apache.plc4x.nifi.util.Plc4xCommon`:
* Reduced repeated code
* Updated mapping table in readme 
* Fixed: reading PlcList not working
* Fixed: PlcByte to Avro mapping was set to bool, caused exception, changed to bytes


plc4j/integrations/apache-nifi/README.md
```diff
@@ -85,29 +85,25 @@ Also, it is important to keep in mind the Processor Scheduling Configuration. Us
...
 Table of data mapping between plc data and Avro
 
+
 | PLC type | Avro Type |
-|----------|-----------|
-| PlcBigDecimal | float |
-| PlcBigInteger | long |
-| PlcBitString | string |
+|----------:|-----------|
 | PlcBOOL | boolean |
```
ottobackwards, line 91: should be byte[]?

ottobackwards (changes requested):
Thanks for the contribution!  This is really nice, only the one comment on the type in the doc being 'bytes'

plc4j/integrations/apache-nifi/README.md
QuanticPony: In Avro specification (https://avro.apache.org/docs/1.11.1/specification/#primitive-types) the primitive type is called bytes

ottobackwards (approved):
Thanks, sorry for my misunderstanding.

Issue #648, opened 2022-11-09
roll back configurate dynamic properties


ottlukas:
Nifi integration has been pleasantly addressed in version 0.10.
Adding a record writer to the source processor is golden.

Unfortunately, the configuration has been changed, and the values to be read must be completed with dynamic properties.
In the previous version, there was a text field (PLC resource address String) where you could simply complete all the addresses to be read out.

Would it be possible to reverse this?  I have a project where I need to read out a lot of values via S7, adding this in NiFi now is endless work.

Imported from Jira [PLC4X-357](https://issues.apache.org/jira/browse/PLC4X-357). Original Jira may contain additional context.
Reported by: lintermans.

chrisdutz:
As the referenced ticket has been merged, can this issue be closed?

QuanticPony:
@chrisdutz It can be closed

Issue #593, opened 2022-11-09
Apache NiFi integration should allow Expression Language


ottlukas:
Apache NiFi integration should allow us to use Expression Language for PLC connection string and PLC resource address string.

We sometimes need to get data from 100 different addresses. Current processors doesn't allow us to create those strings on the fly. So we need to enter them manually or we need utilize NiFi API to create flows automatically. 

If those parameters allows us to use expression language, we can read a list from csv file or from database and then we can read them in a loop in Apache NiFi.  

So it will be very handy feature to dynamically specify the connection string and the address string.

For starting point PutFile processor can be examined. This processor utilizes expression language for it's "Directory" parameter.

[https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java](https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-standard-bundle/nifi-standard-processors/src/main/java/org/apache/nifi/processors/standard/PutFile.java)

Notable Lines:

import org.apache.nifi.expression.ExpressionLanguageScope;

...

Define the parameter as expression language supported.

public static final PropertyDescriptor DIRECTORY = new PropertyDescriptor.Builder()
 .name("Directory")
 .description("The directory to which files should be written. You may use expression language such as /aa/bb/${path}")
 .required(true)
 .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
 *.expressionLanguageSupported(ExpressionLanguageScope.FLOWFILE_ATTRIBUTES)*
 .build();

...

And then script can be evaluated in the onTrigger event like

context.getProperty(DIRECTORY).evaluateAttributeExpressions(flowFile).getValue()

Imported from Jira [PLC4X-196](https://issues.apache.org/jira/browse/PLC4X-196). Original Jira may contain additional context.
Reported by: turker.tunali.

PR #755, opened 2023-01-19
Feature/nifi integration address text

QuanticPony:
Proposal to add a property in NiFi-integration processors to allow users to choose address definition strategy and add support for Expression Language in addresses definition.

## Issues related:
* #648: As stated, adding a lot of addresses as dynamic properties can be tedious. 
* #593 (Jira [PLC4X-196](https://issues.apache.org/jira/browse/PLC4X-196)): Expression Language can be supported easily on addresses definition.

## Actual behavior:
Addresses are specified one by one as dynamic properties in the processor and stored in a map when `onScheduled` method is triggered.
Expression Language is not supported.

## Proposed behavior:
Let user decide between 2 strategies to obtain addresses in the processor:
* Properties as addresses (as actual behavior but with Expression Language): 1 address per dynamic property.
* Property "Address Text": a property where all addresses can be specified as field-value pairs in a JSON.

An address map created every time when `onTrigger` method is called.


ottobackwards:
Thanks for the contribution!
The issues with expression language for addresses and having multiple addresses in that property were at the time:

- Putting something complicated to write in a text field is poor for the user
- The original didn't use json, but used special character substitution changing the native addresses which was a no-go
- The addresses / returns need to have their PLC types resolved, and if the address could change _every_ call ( which you have to assume if you use expression language ) then it would be good to have some kind of caching so that it wasn't so costly every time, and the original PR did not provide that

*In order for this PR to land, it would have to address the caching issue*.  Having json addresses the "don't invent a custom addressing delimiter" issue.  We would just have to live with how poor writing complicated json in that UI would be ( or doing it external and cut and pasting it in)

QuanticPony:
@ottobackwards  regarding the cache:
I have added a cache to store addresses-to-PlcTags and addresses-to-RecordSchema mappings.

I have done some tests and seems to speed up ut to 10x the processors.

At the moment each processor has its cache and a new property: *Schema Cache Size* with default value 1.

Waiting for feedback

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
```diff
@@ -47,9 +47,9 @@ public class Plc4xReadResponseRecordSet implements RecordSet, Closeable {
```
ottobackwards, line 47: Are these TODO's still valid?

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
```diff
@@ -58,9 +58,10 @@ public Plc4xReadResponseRecordSet(final PlcReadResponse readResponse) throws IOE
         rsColumnNames = responseDataStructure.keySet();
                
         if (recordSchema == null) {
```
ottobackwards, line 58: Could we not cache here as well?

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/SchemaCache.java
```diff
@@ -0,0 +1,105 @@
...
+    public SchemaCache(int cacheSize) {
+        this.cacheSize.set(cacheSize);
+    }
+
+    public void setCacheSize(int cacheSize) {
+        this.cacheSize.set(cacheSize);
+        schemaAppendOrder = new AtomicReferenceArray<>(cacheSize);
+        schemaMap = new ConcurrentHashMap<>();
+    }
+
```
ottobackwards, line 55: We need tests for this class to ensure it works properly.
Some JavaDoc would be nice here too.

Also it would be nice if there were some logging here or elsewhere of cache misses etc, in DEBUG mode/level

ottobackwards (changes requested):

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
QuanticPony: With the cache they would not be

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
QuanticPony: This part is already cached on the Plc4xSourceRecordProcessor L187 (https://github.com/zylklab/plc4x/blob/4871a0ca4fd3befc773fd624a33ef31372647751/plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/Plc4xSourceRecordProcessor.java#L187)

plc4j/integrations/apache-nifi/nifi-plc4x-processors/src/main/java/org/apache/plc4x/nifi/record/Plc4xReadResponseRecordSet.java
ottobackwards: Can you remove them please?

QuanticPony:
Thank you for your feedback @ottobackwards!
I have made commits addressing all suggested changes

ottobackwards:
This looks good.  If you are ready, I will let you move it from Draft to ready for review, then I'll finish up

ottobackwards (approved):

hutcheb:
Just going through and removing jackson being pinned at version 2.14.1, this will result in it being bumped to 2.14.2.

This was causing an issue with the build. Let me know if there was a specific reason this was being held back.

QuanticPony:
Must be a left over from testing. There is not any reason, sorry for the inconvenience

PR #776, opened 2023-02-01
Issue 701

glcj:
hi,

Implements the fix for issue-701.
The driver did not allow writing binaries on S7-400 computers.
For the S7-400 it is not required that the last writing element generate the padding. The S7-1500 for example does not take it into account.


chrisdutz:
What's the state of this?

PR #784, opened 2023-02-04
Various Changes

chrisdutz:
This is a Pull Request containing various changes from the repo https://github.com/spnettec/plc4x

It seems they deal with:
- String handling in S7
- Support for S7 200
- Extended syntax of tags to add "encoding" to string fields
- Changes in the Timeout-Handling of our driver base
- Extensions to the OPC-UA driver


chrisdutz:
Admittedly I would not like to merge this PR ... but I think we should have a look at it and possibly pick individual fixes, if we think they are good.

hutcheb:
With the OPCUA protocol package, it seems to include MIT licensed files we use to generate the mspec. While it is a good idea to include them so we don't pull new updates automatically from their Github repo. I don't know if it is compatible with the ASF's rules?

The changes to the OPCUA driver and opcuaserver packages seem to be minimal, but should be good additions.

chrisdutz:
I had a look at what he did. He downloaded the files, added them and added data points that we didn't have. Thought worth investigating what they are?

sruehl:
> Admittedly I would not like to merge this PR ... but I think we should have a look at it and possibly pick individual fixes, if we think they are good.

I set it to draft so we don't do by accident

heyoulin:
@chrisdutz Thank you for this pull request. I have no time to merge my changes. Since I've been so busy recently. I don't have time to do this. Thank you so much

heyoulin:
You shoud take look plc4x driver and server. I modify the tag and can use fit for general read and write tag format

sruehl:
I'm a little bit confused about the commits in this PR. Is this some rebase commit manipulating issue? E.g. why does it contain dependabot commits?

chrisdutz:
I created it by checking out our version at the last version that was merged into the other repo and the other two in its latest version. Then I had intellij compare the two directories and manually went through that huge diff, ignoring generated code and I pulled what I think we should discuss adopting (didn't pull everything)... Hope that explains it.

hutcheb:
Marking this for deletion unless there’s a request to keep it.

PR #795, opened 2023-02-11
Fix code gen concurrent modification

hutcheb:
Small fix for concurrent modification error when validating multiple mspecs.


hutcheb:
Bit of a hack, but it seems to work using a CopyOnWriteArray instead of linkedlist. Would be nice to figure out why a CoModificatioException is reaised otherwise.

Seems to have something to do with mspec types referencing other mspec types in other files.

sruehl:
we might have other issues now like disconnected types etc... It might be better do synchronize and use the example from bacnet to understand why this is happening. So in the bacnet case you can reproduce it by moving up the statement how it was described in the deleted comment. For now we fixed the symptoms but not the root case.

hutcheb:
I agree, certainly need to revisit it. I should have some more time this week to get my head around how it works.

hutcheb:
Hopefully this should be a bit better,
https://github.com/apache/plc4x/commit/e369b2ea12e3b0a863cc3b0b28ede33d45dae18d

I found that when the setter.accept was being called it would occasionally modify the typedefinition list. Running through the consumers until we have nothing more to process seems to work.

PR #796, opened 2023-02-11
fix(plc4j/connection-cache): fix issue with timing of thread

hutcheb:
This test seems to fail every now and then. Hopefully this'll fix it.


chrisdutz:
Yeah... Looks good to me... Thanks for this.

chrisdutz (approved):

PR #818, opened 2023-02-22
New various Changes for #1

spnettec:
fix OutOfMemoryError
add Timeout-Handling
fix connection-cache

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
@@ -63,6 +63,10 @@ public interface PlcTag {
         return PlcValueType.NULL;
     }
 
+    @JsonIgnore
+    default void setPlcValueType(PlcValueType plcValueType){
+    }
+
     /**
      * Returns the number of elements to expect of the response one can expect from this field.
      *
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
@@ -63,10 +63,6 @@ public interface PlcTag {
         return PlcValueType.NULL;
     }
 
-    @JsonIgnore
-    default void setPlcValueType(PlcValueType plcValueType){
-    }
-
     /**
      * Returns the number of elements to expect of the response one can expect from this field.
      *
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
@@ -27,7 +27,6 @@ import org.apache.plc4x.java.api.exceptions.PlcRuntimeException;
 import org.apache.plc4x.java.api.messages.PlcWriteRequest;
 import org.apache.plc4x.java.api.messages.PlcWriteResponse;
 import org.apache.plc4x.java.api.model.PlcTag;
-import org.apache.plc4x.java.api.types.PlcValueType;
 import org.apache.plc4x.java.spi.codegen.WithOption;
 import org.apache.plc4x.java.spi.generation.SerializationException;
 import org.apache.plc4x.java.spi.generation.WriteBuffer;
@@ -196,9 +195,6 @@ public class DefaultPlcWriteRequest implements PlcWriteRequest, Serializable {
                 PlcTag tag = tagValues.getLeft().get();
                 Object[] value = tagValues.getRight();
                 PlcValue plcValue = valueHandler.newPlcValue(tag, value);
-                if (tag.getPlcValueType() == PlcValueType.NULL && value!=null) {
-                    tag.setPlcValueType(plcValue.getPlcValueType());
-                }
                 parsedTags.put(name, new TagValueItem(tag, plcValue));
             });
             return new DefaultPlcWriteRequest(writer, parsedTags);
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
@@ -196,7 +196,7 @@ public class DefaultPlcWriteRequest implements PlcWriteRequest, Serializable {
                 PlcTag tag = tagValues.getLeft().get();
                 Object[] value = tagValues.getRight();
                 PlcValue plcValue = valueHandler.newPlcValue(tag, value);
-                if(tag.getPlcValueType()== PlcValueType.NULL){
+                if (tag.getPlcValueType() == PlcValueType.NULL && value!=null) {
                     tag.setPlcValueType(plcValue.getPlcValueType());
                 }
                 parsedTags.put(name, new TagValueItem(tag, plcValue));
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
@@ -27,6 +27,7 @@ import org.apache.plc4x.java.api.exceptions.PlcRuntimeException;
 import org.apache.plc4x.java.api.messages.PlcWriteRequest;
 import org.apache.plc4x.java.api.messages.PlcWriteResponse;
 import org.apache.plc4x.java.api.model.PlcTag;
+import org.apache.plc4x.java.api.types.PlcValueType;
 import org.apache.plc4x.java.spi.codegen.WithOption;
 import org.apache.plc4x.java.spi.generation.SerializationException;
 import org.apache.plc4x.java.spi.generation.WriteBuffer;
@@ -93,6 +94,7 @@ public class DefaultPlcWriteRequest implements PlcWriteRequest, Serializable {
         return tags.values().stream().map(TagValueItem::getTag).collect(Collectors.toCollection(LinkedList::new));
     }
 
+    @Override
     @JsonIgnore
     public PlcValue getPlcValue(String name) {
         return tags.get(name).getValue();
@@ -194,6 +196,9 @@ public class DefaultPlcWriteRequest implements PlcWriteRequest, Serializable {
                 PlcTag tag = tagValues.getLeft().get();
                 Object[] value = tagValues.getRight();
                 PlcValue plcValue = valueHandler.newPlcValue(tag, value);
+                if(tag.getPlcValueType()== PlcValueType.NULL){
+                    tag.setPlcValueType(plcValue.getPlcValueType());
+                }
                 parsedTags.put(name, new TagValueItem(tag, plcValue));
             });
             return new DefaultPlcWriteRequest(writer, parsedTags);
--- a/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
+++ b/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
@@ -74,8 +74,11 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
                 connectionContainers.put(url, connectionContainer);
             } else {
                 LOG.debug("Reusing exising connection");
-                if(connectionContainer.getRawConnection()!=null && !connectionContainer.getRawConnection().isConnected()){
-                    connectionContainer.getRawConnection().connect();
+                if (connectionContainer.getRawConnection() != null && !connectionContainer.getRawConnection().isConnected()) {
+                    connectionContainer.close();
+                    PlcConnection connection = connectionManager.getConnection(url);
+                    connectionContainer = new ConnectionContainer(connection, maxLeaseTime);
+                    connectionContainers.put(url, connectionContainer);
                 }
             }
         }
--- a/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
+++ b/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
@@ -60,11 +60,6 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
 
     @Override
     public PlcConnection getConnection(String url) throws PlcConnectionException {
-        return getConnection(url,null);
-    }
-
-    @Override
-    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
         ConnectionContainer connectionContainer;
         synchronized (connectionContainers) {
             connectionContainer = connectionContainers.get(url);
@@ -72,13 +67,10 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
                 LOG.debug("Creating new connection");
 
                 // Establish the real connection to the plc
-                PlcConnection connection;
-                if(authentication!=null) {
-                    connection = connectionManager.getConnection(url,authentication);
-                } else{
-                    connection = connectionManager.getConnection(url);
-                }
-                connectionContainer = new ConnectionContainer(connection,maxLeaseTime);
+                PlcConnection connection = connectionManager.getConnection(url);
+
+                // Crate a connection container to manage handling this connection
+                connectionContainer = new ConnectionContainer(connection, maxLeaseTime);
                 connectionContainers.put(url, connectionContainer);
             } else {
                 LOG.debug("Reusing exising connection");
@@ -97,6 +89,11 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
             connectionContainers.remove(url);
             throw new PlcConnectionException("Error acquiring lease for connection", e);
         }
+    }
+
+    @Override
+    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
+        throw new PlcConnectionException("the cached driver manager currently doesn't support authentication");
     }
 
     public void destroy(){
--- a/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
+++ b/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
@@ -58,21 +58,33 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
         this.connectionContainers = new HashMap<>();
     }
 
+    @Override
     public PlcConnection getConnection(String url) throws PlcConnectionException {
+        return getConnection(url,null);
+    }
+
+    @Override
+    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
         ConnectionContainer connectionContainer;
         synchronized (connectionContainers) {
             connectionContainer = connectionContainers.get(url);
-            if (connectionContainers.get(url) == null) {
+            if (connectionContainer == null || connectionContainer.isClosed()) {
                 LOG.debug("Creating new connection");
 
                 // Establish the real connection to the plc
-                PlcConnection connection = connectionManager.getConnection(url);
-
-                // Crate a connection container to manage handling this connection
-                connectionContainer = new ConnectionContainer(connection, maxLeaseTime);
+                PlcConnection connection;
+                if(authentication!=null) {
+                    connection = connectionManager.getConnection(url,authentication);
+                } else{
+                    connection = connectionManager.getConnection(url);
+                }
+                connectionContainer = new ConnectionContainer(connection,maxLeaseTime);
                 connectionContainers.put(url, connectionContainer);
             } else {
                 LOG.debug("Reusing exising connection");
+                if(connectionContainer.getRawConnection()!=null && !connectionContainer.getRawConnection().isConnected()){
+                    connectionContainer.getRawConnection().connect();
+                }
             }
         }
 
@@ -81,14 +93,16 @@ public class CachedPlcConnectionManager implements PlcConnectionManager {
         try {
             return leaseFuture.get(this.maxWaitTime.toMillis(), TimeUnit.MILLISECONDS);
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
+            connectionContainer.close();
+            connectionContainers.remove(url);
             throw new PlcConnectionException("Error acquiring lease for connection", e);
         }
     }
 
-    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
-        throw new PlcConnectionException("the cached driver manager currently doesn't support authentication");
+    public void destroy(){
+        connectionContainers.values().forEach(ConnectionContainer::close);
+        connectionContainers.clear();
     }
-
     public static class Builder {
 
         private final PlcConnectionManager connectionManager;
--- a/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/ConnectionContainer.java
+++ b/plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/ConnectionContainer.java
@@ -65,7 +65,7 @@ class ConnectionContainer {
     public synchronized void close(){
         CompletableFuture<PlcConnection> leaseFuture;
         while((leaseFuture = queue.poll())!=null){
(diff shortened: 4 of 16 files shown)


chrisdutz:
YESS! ... that's much more something we can discuss ... I'll do that as soon as I can (Currently not allowed to do Apache code-work while on the clock :-( ) ... but just wanted to say that it's greatly appreciated :-)

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
```diff
@@ -81,14 +93,16 @@ public PlcConnection getConnection(String url) throws PlcConnectionException {
...
             return leaseFuture.get(this.maxWaitTime.toMillis(), TimeUnit.MILLISECONDS);
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
+            connectionContainer.close();
+            connectionContainers.remove(url);
             throw new PlcConnectionException("Error acquiring lease for connection", e);
         }
     }
 
-    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
-        throw new PlcConnectionException("the cached driver manager currently doesn't support authentication");
```
chrisdutz, line 89: I intentionally left away the autherntication as in general we couldn't have two connections to the same plc with different authentication ... if we do this, we should probably also add the authentication information to the key of the map.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
```diff
@@ -58,21 +58,33 @@ public CachedPlcConnectionManager(PlcConnectionManager connectionManager, Durati
...
+                if(authentication!=null) {
+                    connection = connectionManager.getConnection(url,authentication);
+                } else{
+                    connection = connectionManager.getConnection(url);
+                }
+                connectionContainer = new ConnectionContainer(connection,maxLeaseTime);
                 connectionContainers.put(url, connectionContainer);
             } else {
                 LOG.debug("Reusing exising connection");
+                if(connectionContainer.getRawConnection()!=null && !connectionContainer.getRawConnection().isConnected()){
```
chrisdutz: Instead of exposing the raw connection, I think it might be better to create a new container and not expose the inner connection.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -66,7 +76,7 @@ public void connect() throws PlcConnectionException {
     @Override
     public boolean isConnected() {
         if(connection == null) {
-            throw new PlcRuntimeException("Error using leased connection after returning it to the cache.");
+            return false;
```
chrisdutz, line 81: Not quite sure this is the best way to do this. Because for me there's a difference, if we're not connected or if we already gave back the handle. The way I implemented it gives the user the explicit feedback, that he's doing something bad. If we return false, he might think he just needs to re-connect.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -92,15 +102,139 @@ public PlcReadRequest.Builder readRequestBuilder() {
         if(connection == null) {
             throw new PlcRuntimeException("Error using leased connection after returning it to the cache.");
         }
-        return connection.readRequestBuilder();
+        final PlcReadRequest.Builder innerBuilder = connection.readRequestBuilder();
```
chrisdutz, line 111: Could you please explain what this is needed for?

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
```diff
@@ -194,6 +196,9 @@ public PlcWriteRequest build() {
                 PlcTag tag = tagValues.getLeft().get();
                 Object[] value = tagValues.getRight();
                 PlcValue plcValue = valueHandler.newPlcValue(tag, value);
+                if(tag.getPlcValueType()== PlcValueType.NULL){
```
chrisdutz: As far as I understood it PlcValueType.NULL is used in cases where a value can be null (Like in pointers etc.) ... so it seems as only if you set the type according to the plcValue, if this is NULL ... should not be confused with "null".

chrisdutz (commented):
How do the changes from the Nett wrapper correspond to the ones in the connection-cache?

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
chrisdutz: Also admittedly ... I'm not 100% sure if all drivers safely allow reconnecting if they were somehow disconnected.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
spnettec: This is ready for pl4x driver

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: This is for handling  request failure and close the raw connection

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: If we lose the connection we can handle reconnection

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
spnettec: Sometimes the row connection will die but we can't detected.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
spnettec: Just leave it for the future to authorize

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: By the way. I'm working on eclipse kura. It require work uninterrupted. We must have explicit reconect signal to reconnect not a exception.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
spnettec: By the way. The value can be set null not PlcValueType be set NULL. PLC4x driver write request must give explicit data type. This limits all the driver write tag format for the PLC4x

spnettec: You can checkout my repo and take look PL4x driver and PL4x server

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: I have already explained that previous PR are the basis for subsequent PRs. Next I want to submit a PR for PL4x

spnettec:
The raw conection is netty conection and if disconnected it always can
reconnect

Christofer Dutz ***@***.***> 于2023年2月22日周三 22:57写道：

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
hongjinlin: > 



> 

I have encountered a scenario where the PLC connection appeared to be disconnected, but in reality it was still connected due to certain reasons. However, we kept trying to reconnect and as a result, the TCP estab value increased to tens of thousands. This caused the PLC to be disconnected and also resulted in server network issues. Therefore, I'm looking to see if there is a better solution available.

hongjinlin: > I have already explained that previous PR are the basis for subsequent PRs. Next I want to submit a PR for PL4x

First of all, thank you for your PR. Secondly, I think it would be beneficial if you could discuss your entire design approach with everyone, rather than just submitting a basic PR and following up with subsequent PRs.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
chrisdutz: Well I'd rather simply do it right, right away ... shouldn't be too much additional work ... sort of like a "url + authentication.toString()"

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
chrisdutz: I still don't quite understand the problem with this ... if you returned the connection to the pool, you should not do anything with it ... that's sort of the contract on using the pool. If there's some logic calling "isConnected()" after it has been returned, that's just a flaw in the program using it. I still think we should continue to throw an exception here.

plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
```diff
@@ -63,6 +63,10 @@ default PlcValueType getPlcValueType() {
         return PlcValueType.NULL;
     }
 
+    @JsonIgnore
+    default void setPlcValueType(PlcValueType plcValueType){
+    }
```
chrisdutz: Hmm ... this adds a default implementation, which doesn't really do anything. Also do we usually implement all of the types in the API module, to create immutable types. Is there any reason for this addition? I think our tags should generally stay immutable types.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
```diff
@@ -194,6 +196,9 @@ public PlcWriteRequest build() {
                 PlcTag tag = tagValues.getLeft().get();
                 Object[] value = tagValues.getRight();
                 PlcValue plcValue = valueHandler.newPlcValue(tag, value);
+                if (tag.getPlcValueType() == PlcValueType.NULL && value!=null) {
+                    tag.setPlcValueType(plcValue.getPlcValueType());
+                }
```
chrisdutz: I still don't think we should have this in here ... if the tag says it's one type, but the PlcValue says it's something else, we shouldn't update this (I guess this is the reason you needed to add the "setPlcValue"). We should fire an exception, because in this case the user is doing something incorrectly.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
```diff
@@ -58,21 +58,33 @@ public CachedPlcConnectionManager(PlcConnectionManager connectionManager, Durati
...
+    @Override
     public PlcConnection getConnection(String url) throws PlcConnectionException {
+        return getConnection(url,null);
+    }
+
+    @Override
+    public PlcConnection getConnection(String url, PlcAuthentication authentication) throws PlcConnectionException {
         ConnectionContainer connectionContainer;
         synchronized (connectionContainers) {
             connectionContainer = connectionContainers.get(url);
```
chrisdutz, line 65: As this now includes the authentication, we should probably have a more generic "getContainerId" method, that takes a url and authentication and if authentication is not null, then it sort of does a "url + authentication.toString()" to allow multiple connections with different users (I could imagine this being a topic in OPC-UA)

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/ConnectionContainer.java
```diff
@@ -28,12 +28,20 @@
...
+    private boolean closed = false;
     private final Duration maxLeaseTime;
     private final Queue<CompletableFuture<PlcConnection>> queue;
 
     private LeasedPlcConnection leasedConnection;
 
+    public boolean isClosed() {
+        return closed;
+    }
+    public PlcConnection getRawConnection() {
```
chrisdutz, line 41: In contrast to your comment, this does actually return the Plc4x connection and not the Netty connection ... we shouldn't expose this, as this allows "stealing" the connection.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/ConnectionContainer.java
```diff
@@ -54,8 +62,19 @@ public synchronized Future<PlcConnection> lease() {
         }
         return connectionFuture;
     }
-
+    public synchronized void close(){
+        CompletableFuture<PlcConnection> leaseFuture;
+        while((leaseFuture = queue.poll())!=null){
+            leaseFuture.complete(null);
```
chrisdutz: Wouldn't it make more sense to complete exceptionally instead of returning null? We are currently expecting the result of a getConnection() to be a usable connection or an exception ... returning "null" sort of doesn't match that contract. I would change it to "completeExceptionally"

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -92,15 +102,139 @@ public PlcReadRequest.Builder readRequestBuilder() {
...
+            public PlcReadRequest build() {
+                final PlcReadRequest innerPlcReadRequest = innerBuilder.build();
+                return new PlcReadRequest(){
+
+                    @Override
+                    public CompletableFuture<? extends PlcReadResponse> execute() {
+                        CompletableFuture<? extends PlcReadResponse> future = innerPlcReadRequest.execute();
+                        final CompletableFuture<PlcReadResponse> responseFuture = new CompletableFuture<>();
+                        future.handle((plcReadResponse, throwable) -> {
+                            if (plcReadResponse != null) {
```
chrisdutz: Could we change this to "throwable == null"? ... it generally means the same, but it more sets the focus on "nothing went wrong".

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -92,15 +102,139 @@ public PlcReadRequest.Builder readRequestBuilder() {
...
+                                responseFuture.complete(plcReadResponse);
+                            } else {
+                                try {
+                                    destroy();
+                                } catch (Exception e) {
+                                }
+                                responseFuture.completeExceptionally(throwable);
```
chrisdutz, line 131: So do I understand this block correctly? If anything goes wrong with the request, it terminates the connection, right? Should we really be doing this? I mean ... it's easy to ask for an invalid tag ... that doesn't really make it necessary to reconnect? For example with my KNX gateway it only allows 3 connections in paralell and within a window of 2-3 minutes (when it closes inactive connections) ... not 100% sure we should be doing this.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -51,13 +51,23 @@ public void run() {
 
     @Override
     public synchronized void close() {
-        // Make the connection unusable.
-        connection = null;
-
+        if(connectionContainer == null) {
+            return;
+        }
```
chrisdutz, line 58: Why should the connectionContainer ever be null? The way I understand it, the only way this could happen, would be that a program got a lease ... executed a request, that had an issue and therefore it automatically closed the connection and destroy() called close from the inside and then the using application calls close regularly ... not too happy with this workflow in general ... we should really discuss things on the list.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
chrisdutz: But I guess ... otherwise there's no way for the user to actually invalidate a connection ... I guess we should probably have different sets of Exceptions that we return. Some that indicate the connection is no longer usable and some that say: you did something wrong, please do it right next time (Like invalid addresses)

plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
chrisdutz: This seems to be related to the change in DefaultPlcWrite request, were the plc value type can be updated ... as I don't really like that change, I think we should also remove this change ... as both actually don't have anything to do with what we would like to address with this PR

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
spnettec: image

I want to change plc4x tagAddress can praser other driver. eg:  %DB2:94:STRING(10), %DB2:78:LREAL[2]

spnettec: It can remove tag formate that depends on other driver's label formats.
I can remove this change. Next PR. I want to fix plc4x driver and server

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
spnettec: I never use authentication. Just follow the api. I can remove this change

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/ConnectionContainer.java
spnettec: I agree that this should throw exception

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: Right. Thank you

plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
spnettec: image

This is ready for PLC4X Driver. Maybe i should change these in next PR

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcWriteRequest.java
chrisdutz: Please remove this from this PR for now as we have to deal with this "feature" separately.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/CachedPlcConnectionManager.java
chrisdutz: Well then we should leave it as it was ... either we support it and then we should really support it, or we don't and we throw an exception (as it was before)

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: Invalid tag is not throw exception. It is normal return result. This is unexcepted exception not caught by protocol. So this error is very serious network error or protocol data error

plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
chrisdutz: Yes please ... as I mentioned, we should try to keep one PR separated from the other ... we're already mixing up the Cache and the Timeouts (which I think is ok as I think both are not controversial)

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
chrisdutz: Ok ... I think you are right with this. So please ignore my comment.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: Close twice. eg: user close this connection but usageTimer will close it again. It will cause nullpointexception

plc4j/api/src/main/java/org/apache/plc4x/java/api/model/PlcTag.java
spnettec: Ok. I will rollback

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: We use PLC4x to collect data for long term work. If the connection is closed abnormally, I don’t want to always use the try catch method. If we use this to detect that the connection is closed, we only need to borrow it again.

chrisdutz: I would say that's a usecase particular to how you use PLC4X, however out API shouldn't be optimized towards one particular usecase. I think try-catch-blocks don't consume any noticable amount of resources. So I would like to ask to to stick to using try-catch blocks and us keeping the API coherent.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
chrisdutz: Yeah ... when returning a connection after a connection timed-out or got another error, it would have thrown an error from inside the returnConnection method of the Container.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: Just connect is not allowed. I think isConeccted is should reserve the functionality of the original API

spnettec: I'm wondering why you want to disable this API functionality and replace it with an error

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
spnettec: This is a bug.  It should not throw an error. If you change usageTimer from local variable to private variable i can canel the Time. But...  I still think the Timer is not necessary. So i don't do anyting. Just add this code.

Issue #821, opened 2023-02-23
Reorganize timeout handling in plc4j-spi


splatch:
### What would you like to happen?

The timeout handling in plc4j is affected by its own logic where registered handlers are called only when packets arrive. This means that in some cases, when communication line is broken or simply hanged, timeouts may never be reported back to driver.

There were several attempts to fix that, however none of them have worked. The recent work in #818 shed some light on some of netty utilities which could be wired into the spi to separate timeout handling. Given that our use of netty in plc4j is never ending source of troubles and confusion (it orchestrates a lot!), we may consider introducing new SPI types to hide specifics, even between our own internals.

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] S7

splatch:
Addressed through #822.

spnettec:
Good job. @splatch

PR #822, opened 2023-02-23
feat(plc4j) Better handling of timeouts in plc4j (#821).

splatch:
The tricky part I was not able to solve yet is how to wire custom timeout manager into the driver. The configuration instances we have are not well defined and passing them around is not clear to me.

### DIFF
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
@@ -23,23 +23,26 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.MessageToMessageCodec;
 import io.vavr.control.Either;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.plc4x.java.api.authentication.PlcAuthentication;
+import org.apache.plc4x.java.spi.TimeoutManager.CompletionCallback;
 import org.apache.plc4x.java.spi.configuration.Configuration;
 import org.apache.plc4x.java.spi.events.*;
+import org.apache.plc4x.java.spi.internal.DefaultConversationContext;
 import org.apache.plc4x.java.spi.internal.DefaultExpectRequestContext;
 import org.apache.plc4x.java.spi.internal.DefaultSendRequestContext;
 import org.apache.plc4x.java.spi.internal.HandlerRegistration;
+import org.apache.plc4x.java.spi.netty.NettyHashTimerTimeoutManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
-import java.time.Instant;
 import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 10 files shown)


spnettec:
> The tricky part I was not able to solve yet is how to wire custom timeout manager into the driver. The configuration instances we have are not well defined and passing them around is not clear to me.

I think we can config in the driver

```
    @ConfigurationParameter("timeout-request")
    @IntDefaultValue(4000)
    protected int timeoutRequest;

    @Override
    public TimeoutManager getTimeoutManager() {
        return new TimeoutManager(){

            @Override
            public CompletionCallback register(TimedOperation operation) {
                // return null;
                // here do something
            }

            @Override
            public void stop() {
                 // here do something
            }
        };
    }
```

And send request like this

```
transaction.submit(() -> context.sendRequest(tpktPacket)
            .onTimeout(new TransactionErrorCallback<>(future, transaction))
            .onError(new TransactionErrorCallback<>(future, transaction))
            .expectResponse(TPKTPacket.class, Duration.ofMillis(configuration.getTimeoutRequest()))
            .check(p -> p.getPayload() instanceof COTPPacketData)
            .unwrap(p -> ((COTPPacketData) p.getPayload()))
            .unwrap(COTPPacket::getPayload)
            .check(p -> p.getTpduReference() == tpduId)
```

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -200,85 +198,60 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
+            this.protocolBase.onConnect(new DefaultConversationContext<>(this::registerHandler, ctx, authentication, passive));
         } else if (evt instanceof DisconnectEvent) {
-            this.protocolBase.onDisconnect(new DefaultConversationContext<>(ctx, authentication, passive));
+            this.protocolBase.onDisconnect(new DefaultConversationContext<>(this::registerHandler, ctx, authentication, passive));
         } else if (evt instanceof DiscoverEvent) {
-            this.protocolBase.onDiscover(new DefaultConversationContext<>(ctx, authentication, passive));
+            this.protocolBase.onDiscover(new DefaultConversationContext<>(this::registerHandler, ctx, authentication, passive));
         } else if (evt instanceof CloseConnectionEvent) {
-            this.protocolBase.close(new DefaultConversationContext<>(ctx, authentication, passive));
+            this.protocolBase.close(new DefaultConversationContext<>(this::registerHandler, ctx, authentication, passive));
```
spnettec, line 205: Here we should stop TimeManager

splatch: You're right, however I am not entirely sure of netty vs plc4x lifecycle. Namely can we reestablish connection  once we reached that point? At high level plc4x drivers support `connect` call, we would need to assure that it  always configure fresh pipeline and uses new timeout manager forcing netty to fire bootstrap procedure again.

splatch (commented):
Thanks for initial review, were you able to test this PR against your environment?

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
spnettec: Fresh pipeline finnaly always invoke `PlcConnectionManager.getConnection` it will create fesh pipeline. If you don't stop the TimeManager it will caouse memory leak

spnettec:
> Thanks for initial review, were you able to test this PR against your environment?

What's you changed?

splatch:
This PR was open for quite long time. I'm going to address @spnettec review comments in separate change set.

PR #832, opened 2023-03-07
Fix reading UTF-8 strings (from OPC UA nodes)

Planet-X:
When reading string values via the OPC UA protocol a `StringIndexOutOfBoundsException` was thrown whenever the string contained non-ASCII characters. Upon further inspection, the problem could be pinned down to the logic in `readString()` of`ReadBufferByteBased`.

The `readString()` function in `ReadBufferByteBased` could result in a `StringIndexOutOfBoundsException` upon calling `substring()`. Reason is that the calculated `realLength` is in bytes and the length of the created string is the number of characters. For UTF encodings, multiple bytes can be just one character. This causes `realLength` to be longer than the actual string and thus the `substring()`-call to fail.

This PR fixes the issue by applying `realLength` to the byte-length instead of the string length: The byte array is sliced to length `realLength` and then converted to the final string. `substring()` isn't used anymore.
The fix can be verified via the included regression test and has additionally been tested locally using the OPC UA protocol.
Maybe other protocols were also affected?

Additionally I've included some other minor fixes for the logic. Check the commits for these.

Regards,
Marc


sruehl (approved):
Thanks for the PR: LGTM, although @chrisdutz should check the ASCII part regarding the padding

PR #839, opened 2023-03-11
Feat/hop

glcj:
Integration with Hop V2.X.X in progress. Action and Transform Read, Write and Events are working, Subscription is missing due to problems in the S7 driver, these routines were removed in version 0.10.0 of Plc4x. The resources for the modification are added to the portal and proceed to their documentation.


chrisdutz:
Your mspec changes shouldn't cause any serious problems ... I'll have a look at if the build works on my machine.

I think you can remove the ".cache" files as I think IntelliJ creates them when using the IDE editor ... same with the generated png files. These should be created during the website generation.

I love your Toddies by the way.

chrisdutz:
What's the status of this PR?

PR #962, opened 2023-05-27
Add support of PlcDINT for BigInteger in PlcValueHandler

PatrykGala:
1. The uncaught `PlcUnsupportedDataTypeException` from `PlcValueHandler` caused the CompletableFuture not to complete, the `OpcuaPlcDriverTest` hangs on `CompletableFuture.get()`
2. The root of the problem is the missing support of BigInteger in [PlcValueHandler.java](https://github.com/apache/plc4x/compare/develop...PatrykGala:plc4x:fix-test#diff-4a6f80d3ad2ced6c8c8bf3f146280004412c868a26cb1539b68486dce2671ba3)

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
@@ -195,7 +195,7 @@ public class OpcuaProtocolLogic extends Plc4xProtocolBase<OpcuaAPU> implements H
                         future.complete(new DefaultPlcReadResponse(request, status));
                         return;
                     }
-                } catch (ParseException e) {
+                } catch (Exception e) {
                     future.completeExceptionally(new PlcRuntimeException(e));
                 }
             };
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
@@ -195,7 +195,7 @@ public class OpcuaProtocolLogic extends Plc4xProtocolBase<OpcuaAPU> implements H
                         future.complete(new DefaultPlcReadResponse(request, status));
                         return;
                     }
-                } catch (Exception e) {
+                } catch (ParseException|PlcRuntimeException e) {
                     future.completeExceptionally(new PlcRuntimeException(e));
                 }
             };
--- a/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
+++ b/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
@@ -43,7 +43,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.concurrent.ExecutionException;
 
-@Disabled("Currently seems to block")
 public class OpcuaPlcDriverTest {
 
     @BeforeAll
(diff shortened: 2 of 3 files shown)


plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
chrisdutz, line 1: Please don't just disable tests without opening a new issue for it ... this way we have already disabled very many tests, tests we don't know about or forget.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/values/PlcValueHandler.java
```diff
@@ -83,6 +83,9 @@ public static PlcValue of(Object[] values) {
         if (value instanceof Long) {
             return PlcLINT.of(value);
         }
+        if (value instanceof BigInteger) {
+            return PlcDINT.of(value);
```
chrisdutz: Woudln't a LINT be the better match, as a BigInteger could actually be even bigger than the DINT value space?

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
```diff
@@ -195,7 +195,7 @@ public CompletableFuture<PlcReadResponse> read(PlcReadRequest readRequest) {
                         future.complete(new DefaultPlcReadResponse(request, status));
                         return;
                     }
-                } catch (ParseException e) {
+                } catch (Exception e) {
```
chrisdutz: What was the reasoning behind this?

chrisdutz (changes requested):
Please don't disable tests ... the test isn't blocking on my computer or on the CI servers, so if your changes make it block, we probably should take care of this.

plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
PatrykGala: I am enabling in this PR, the test was disabled in https://github.com/apache/plc4x/commit/0af938f7be72cb6e1a3091ff589ee4fb20436928

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/values/PlcValueHandler.java
PatrykGala: Yes, PlcLINT is better choose. I've pushed the change.

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
PatrykGala: The uncaught `PlcUnsupportedDataTypeException` thrown by `PlcValueHandler` prevents the `CompletableFuture` from completing, causing the thread to hang on `CompletableFuture.get()`. By making this change, the test will now complete with an Exception instead of hanging indefinitely.

splatch:
@chrisdutz if you're ok with changes (PlcLINT is in place) I'd proceed with merging.

chrisdutz:
Please enable the test first

splatch:
> Please enable the test first

As far I can see this PR enables test which was disabled.

chrisdutz:
Indeed it does... So ignore my mumbling 😁

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/protocol/OpcuaProtocolLogic.java
sruehl:
```suggestion
                } catch (ParseException|PlcRuntimeException e) {
```
This would be a better approach then (e.g Exception is also catching InterruptedException)

PR #1028, opened 2023-07-11
fix(plc4go/modbus): Delete elements in the loop, and the index is dec…

hongjinlin:
In some batch read/write cases, because the element is deleted in the loop, but the index is not decremented by 1, the program freezes, and even the timeout does not work.

### DIFF
--- a/plc4go/spi/default/DefaultCodec.go
+++ b/plc4go/spi/default/DefaultCodec.go
@@ -268,6 +268,7 @@ func (m *defaultCodec) HandleMessages(message spi.Message) bool {
 				}
 				continue
 			}
+			m.log.Trace().Msg("message handled")
 			messageHandled = true
 			// If this is the last element of the list remove it differently than if it's before that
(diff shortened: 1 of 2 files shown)


sruehl (changes requested):
Please add a unit test for this change. It is such a central change which needs a test to proof the failure before the change and that it is working right after the change. The test then I can try out on the current develop to understand what exactly is happening

sruehl:
Small note: you can add a testcase in `DefaultCodec_test.go` before line 686

sruehl:
And looking at it: And easy fix would be to break the loop after we found someone who accepted the message. There should be only one at a time accepting the message I think

sruehl (approved):
I could replicate the problem and verify that your fix fixes it. Thanks for the PR 🎉

hongjinlin:
> I could replicate the problem and verify that your fix fixes it. Thanks for the PR 🎉

Thanks for reviewing and adding testcase :sunglasses: .

PR #1036, opened 2023-07-23
feat: First draft of a publication API (Still needs some fine-tuning)

chrisdutz:
Initial version of a publication API, which is intended to mirror the subscription API.
It is currently only needed for the coming Profinet NG driver, which allows not only subscribing, but also publishing data cyclically.


chrisdutz:
Admittedly I'm not 100% happy with our current Subscription part ... but I'd currently try to keep the Publication API in line with the current Subscription API.

splatch:
Publication API is a good idea. I remember that I had to implement some workarounds for canopen above plc4x api, just to keep other bus participants happy. Also, some tools/chips may have their way to handle "publish" tasks due to time requirements - from this point of view having a higher level API to draw contract for applications is definitely a step forward!

sruehl:
is this still relevant?

chrisdutz:
It would only be relevant for working drivers such as Profinet or EtherCat, which we don't have.

splatch:
Given that we do not have immediate need for such API we can postpone this until driver or drivers with such requirement appear. As for now, end user can always use separate thread which will write specific tags application requires to work.

PR #1088, opened 2023-08-29
Feature/iec 60870 5 104

chrisdutz:
So ... I've implemented the small change on adding a map of Metadata to the PlcValues (Per default they simply contain a reference to the emptyMap(), but as soon as someone calls addMetaData, it will pass along stuff. 

Feel free to inspect the Iec608705104TagParser type for how to use it.


chrisdutz:
So ... I guess I'll be merging this tomorrow, if I don't hear any opposition.

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValue.java
```diff
@@ -149,4 +149,10 @@ public interface PlcValue {
 
     Map<String, ? extends PlcValue> getStruct();
 
+    Map<String, PlcValue> getMetaData();
+
+    boolean hasMetaData(String key);
```
sruehl, line 154: not sure if those are worth it if you expose a map anyway. This just bloats the API

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValue.java
```diff
@@ -149,4 +149,10 @@ public interface PlcValue {
...
+
+    PlcValue getMetaData(String key);
```
sruehl, line 156: not sure if those are worth it if you expose a map anyway. This just bloats the API

sruehl (changes requested):

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/values/PlcNull.java
```diff
@@ -271,6 +271,21 @@ public PlcValue getValue(String key) {
         return null;
     }
 
+    @Override
+    public Map<String, PlcValue> getMetaData() {
+        return null;
```
sruehl: Maybe provide a `collections.emptyMap()` (or however this is called, might be a static variable too) just to avoid NPEs when calling it

sruehl: or if it extends PLC Value adapter remove those implementations here

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValue.java
chrisdutz: Guess I mainly added it, because we did it in the rest of the API.

sruehl (changes requested):

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/values/PlcNull.java
chrisdutz: Fixed

plc4j/api/src/main/java/org/apache/plc4x/java/api/value/PlcValue.java
chrisdutz: I actually removed the raw map access.

PR #1099, opened 2023-09-22
plc4j-driver-opcua: Fix incorrectly handled GUID tags

takraj:
1. The driver tried to `arraycopy()` a `Long`, which caused `ArrayStoreException`.
 2. GUIDs are structured objects, that matters when it comes to byte ordering. The OPC-UA protocol uses little (mixed) endian GUID encoding format, while Java uses the standard big endian one (RFC4122), so conversion is needed.

References:
 * https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.2.6
 * https://en.wikipedia.org/wiki/Universally_unique_identifier#Encoding
 * https://devblogs.microsoft.com/oldnewthing/20220928-00/?p=107221
 * https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html
 * https://www.ietf.org/rfc/rfc4122.txt


sruehl:
Thanks for the PR @takraj 🚀

Issue #1104, opened 2023-09-23
[Feature Request]: plc4j-driver-opcua - Add support for `PlcUsernamePasswordAuthentication`


takraj:
### What would you like to happen?

The OPC-UA driver supports (among others) password based authentication. Currently one can only configure this through URL query parameters, like so:

```
driverManager.getConnection("opcua:tcp://127.0.0.1:4567?username=foo&password=bar")
```

There already is a facility in the framework to pass authentication object to the `getConnection(...)` call, so ideally the above should be equivalent to this one below:

```
driverManager.getConnection(
    "opcua:tcp://127.0.0.1:4567",
    new PlcUsernamePasswordAuthentication("foo", "bar")
)
```

Looking at the code, this feature is currently only supported by `plc4j-driver-ads`, and other drivers silently ignore the passed object.

Adding support for this facility would improve the overall user experience.

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [X] OPC-UA
- [ ] S7

sruehl:
fixed with #1107

PR #1107, opened 2023-09-24
plc4j-driver-opcua: Add support for PlcUsernamePasswordAuthentication

takraj:
Attempts to resolve issue #1104


sruehl:
Thanks for the PR @takraj 🚀

PR #1108, opened 2023-09-24
plc4j-driver-opcua: Adapt error handling of reads, to writes and subscriptions

takraj:
This fixes `ClassCastException` coming in case of errors.

Also covers an earlier draft PR: https://github.com/apache/plc4x/pull/317


sruehl:
Thanks for the PR @takraj 🚀

Issue #900, opened 2023-04-14
[Bug]: plc4j-tools-connection-cache: broken connections remaing in the cache on timeout


QuanticPony:
## Summary
When a connection stored in the connection-cache breaks due to a network failure, the connection is not removed from the cache and blocks future uses of the same connection string.

## Context
Encountered while trying to solve a similar problem as https://github.com/apache/plc4x/issues/623 in the NiFi integration:
When a processor is running and the network connection to the PLC is interrupted, the processors continues to throw errors even if the network connection is restored.

This was brought up in a mail by me (https://lists.apache.org/thread/xm38nh8xzh1m1kj0y74dx0goo81cos82) that sparked a pull request by [heyoulin](https://github.com/spnettec) (https://github.com/apache/plc4x/pull/818), an issue by [splatch](https://github.com/splatch) (https://github.com/apache/plc4x/issues/821) and a commit from [@chrisdutz ](https://github.com/chrisdutz) (https://github.com/apache/plc4x/commit/9b06c2de0c77a7c1bbcb730bb5285c4435002c93).

The commit (https://github.com/apache/plc4x/commit/9b06c2de0c77a7c1bbcb730bb5285c4435002c93) did not fully addressed the problem, so I bring my attempt to fix it.

## Replicate the problem
In order to replicate the problem use the code at the end and follow the steps:
1) Start the main below
2) Disconnect network
3) Wait until errors are shown in the stdout
4) You will see the connection is been used after it fails:
```
16:38:22.486 [main] DEBUG o.a.p.j.u.c.CachedPlcConnectionManager.getConnection:72 - Reusing exising connection
Failed to read due to: 
java.util.concurrent.TimeoutExceptio
```
5) Reconnect network. The problem persists.

## Possible Solution
The LeasedConnection returns a Future that encapsulates the Future that connects to the PLC. The second one is the one that can mark the connection as invalid for removal. For the moment I have been able to work around this by overriding the `get` method of the first Future:
```java
@Override
public PlcReadResponse get(long timeout, TimeUnit unit)
        throws InterruptedException, ExecutionException, TimeoutException {
    try {
        return super.get(timeout, unit);
    } catch (TimeoutException e) {
        future.completeExceptionally(e);
        throw e;
    }
}
```
You can see my solution in the zylklab fork (https://github.com/zylklab/plc4x/tree/Fix/nifi-integration-timeout). If you could give me some feedback I would like to make this into a PR as soon as posible.

_____
```java
[...]

PR #1122, opened 2023-09-29
Nifi integration revision

### LINKED ISSUE #593: Apache NiFi integration should allow Expression Language

### LINKED ISSUE #629: Nifi  BasePlc4xProcessor define validation logic for PLC_CONNECTION_STRING and PLC_ADDRESS_STRING properties
Apache Nifi integration, define logic for the validation methods included in the class BasePlc4xProcessor, for the properties PLC_CONNECTION_STRING and PLC_ADDRESS_STRING

Imported from Jira [PLC4X-328](https://issues.apache.org/jira/browse/PLC4X-328). Original Jira may contain additional context.
Reported by: inigoao.

QuanticPony:
# NiFi integration revision
As we aim for a new release I bring a revision of the NiFi integration with a couple of improvements:

* #900: this made the processors be able to auto-reconnect with small changes. And allowed to make:
    * #593 : this was already done for the addresses. But the connection string had to be constant. Expression Language is now supported on the connection string.
    * #629 : this was an old TODO. I have added validation for the connection string (if not EL in use) and to the addresses (if not EL in use in neither of them as we need the driver).

* Add FilePropertyAccessStrategy: now we can read the addresses from a file in JSON format. Same as 'Address Text' property but in a file. Expression Language can be used and validation for the tags also works.
*  Processors code revision: rewrote the onTrigger methods of the processors so they look more similar between them. To make them more maintainable.
* Complete FlowFlile transfer: some FlowFiles did not transfer to failure correctly. Now the incoming FlowFile is sent to failure and one attribute is added with the localized message in case of any exception.
* Timestamp field name: It was a fixed `ts` field/attribute we added. I have made a property that defines this field's name. 
* Update Readme to reflect changes.

Closes #593
Closes #629

### DIFF
--- a/plc4j/integrations/apache-nifi/README.md
+++ b/plc4j/integrations/apache-nifi/README.md
@@ -20,8 +20,9 @@ under the License.
 
 ## Common properties
 The following properties applies to all Plc4x Processors:
-* Connection String: A constant connection string such as `s7://10.105.143.7:102?remote-rack=0&remote-slot=1&controller-type=S7_1200`.
-* Read/Write/Subscribe timeout (miliseconds): Specifies the time in milliseconds for the connection to return a timeout. In case of subscription the timeout is used to renew connections.
+* Connection String: A constant connection string such as `s7://10.105.143.7:102?remote-rack=0&remote-slot=1&controller-type=S7_1200` or a valid Expression Language ([Expression Language NiFi documentation](https://nifi.apache.org/docs/nifi-docs/html/expression-language-guide.html)) such as `${plc4x.connection_string}`.
+* Timeout (miliseconds): Specifies the time in milliseconds for the connection to return a timeout. Is used to renew connections. Can be set with Expression Language.
+* Timestamp field name: It defines the name of the field that represents the time when the response from the Plc was received. It will be added to the attributes or to the record deppending on the processor used.
 * Address Access Strategy: defines how the processor obtains the PLC addresses. It can take 2 values:
   * **Properties as Addreses:** 
       For each variable, add a new property to the processor where the property name matches the variable name, and the variable value corresponds to the address tag. 
@@ -87,10 +88,6 @@ Table of data mapping between plc data and Avro types (as specified in [Avro spe
 Also, it is important to keep in mind the Processor Scheduling Configuration. Using the parameter **Run Schedule** (for example to *1 sec*), the reading frequency can be set. Note that by default, this value is defined to 0 sec (as fast as possible).
 
 
-## Plc4xSinkProcessor
-
-## Plc4xSourceProcessor
-
 ## Plc4xSinkRecordProcessor
 
 This processor is <ins>record oriented</ins>, reads from a formated input flowfile content using a Record Reader (for further information see [NiFi Documentation](https://nifi.apache.org/docs/nifi-docs/html/record-path-guide.html#overview)). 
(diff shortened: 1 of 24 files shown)


chrisdutz:
Thanks for this ;-) 

Unfortunately I am probably not the one ideal to review NiFi related things, but I would trust you and your changes. So if I don't hear any objections till Monday, I'll merge this PR into develop before cutting the release branch.

chrisdutz:
By the way ... would be cool, if you could sign up to the dev-list and start communicating with us there ;-)

ottobackwards (changes requested):
Thanks for the work!

- There is a lot of duplication of methods between each of the processors.  These should be moved so that we don't have multiple copies in each processor
- I would like to see more tests for the new evaluation enabled things
- I would like to see tests around the caching with evaluations
- having properties read from files is not a great pattern for nifi, and there are no tests for it

in short, please add tests for all the new things and things that are now effected by the permutations possible.  Please de-duplicate the methods in the various processors.

In the future, smaller pr's would make it easier to review

QuanticPony:
Sorry for the long PR. I would normally split it, but though posting as early as possible would be best in this case.
Duplicated code removed. Unit tests added.
i added the read tags from a file as I have seen it in some production environments with other OPCUA processors. I think it could be easier to migrate to PLC4X if we have them. If you don't see it clear i can remove it

QuanticPony:
I'm done. Tell me if anything else is needed. Will have a look at it over the weekend

TurkerTunali:
So, should we try the NiFi integration now?

chrisdutz:
Well it should be part of the Release ... so feel free to try.

PR #1139, opened 2023-10-10
plc4j-driver-opcua: Fix keepalive threads are never shut down

takraj:
Attempts to fix #1138 

* OPC-UA keepalive threads now have an exit condition.
* Both keepalive and subscription loops are now executed in their own, dedicated thread, instead of using the `ForkJoin` common pool.
* Keepalive thread now uses the more sensible session timeout, instead of the 10 hour `CONNECTION_LIFETIME`. The keepalive messages are now actually sent, checked with Wireshark.
* Swapped the sleep & communication logics in the keepalive thread.
* Keepalive responses now increment the `senderSequenceNumber`.
* Smoke tests added to verify the solution.

PR #1147, opened 2023-10-13
plc4j-driver-opcua: Await `writeAndFlush(msg)` & send next msg async

takraj:
All OPC-UA messages contain a sequence number, which is independent from the normal TCP sequence numbers. This sequence number is checked by the server, that may refuse to respond, if the sequence number is lower than the last received message's. Packetization itself happens asynchronously in netty, some time after `writeAndFlush(msg)` has been invoked by the application code. If there are concurrent calls to this function, there is no guarantee that messages are packeted in the same order as they have been added to `writeAndFlush(msg)`. This, in some cases, can cause OPC-UA messages being delivered to the server in different order, than it is specified by their sequence numbers, and the server may drop these messages, and the client eventually times out.

As TCP guarantees delivering packets in the correct order, a trival solution is to simply wait for packetization to complete, before adding the next message to the pipeline.

But yet comes a design problem: OPC-UA response handlers were written in a way, that they may send new messages, while processing the response. These handlers are executed on the netty event loop thread, which is shared among all Netty I/O operations, such as receiving/sending messages. This essentially means, that if a response handler is currently being executed, you cannot start packetizing a new message, because that would require the response handler to finish, which is waiting for guess what, the packetization to complete.

Solution to this is to execute response handlers asynchronously, so they don't occupy the Netty event handler thread.

This commit also fixes the `OpcuaPlcDriverTest` flakyness, experienced after https://github.com/apache/plc4x/pull/1139 had been merged.

See also: https://ci-builds.apache.org/job/PLC4X/job/PLC4X/job/develop/1695/

### DIFF
--- a/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/context/SecureChannel.java
+++ b/plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/context/SecureChannel.java
@@ -20,6 +20,7 @@ package org.apache.plc4x.java.opcua.context;
 
 import static java.lang.Thread.currentThread;
 import static java.util.concurrent.Executors.newSingleThreadExecutor;
+import static java.util.concurrent.ForkJoinPool.commonPool;
 
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.RandomUtils;
@@ -178,7 +179,7 @@ public class SecureChannel {
         }
     }
 
-    public void submit(ConversationContext<OpcuaAPU> context, Consumer<TimeoutException> onTimeout, BiConsumer<OpcuaAPU, Throwable> error, Consumer<byte[]> consumer, WriteBufferByteBased buffer) {
+    public synchronized void submit(ConversationContext<OpcuaAPU> context, Consumer<TimeoutException> onTimeout, BiConsumer<OpcuaAPU, Throwable> error, Consumer<byte[]> consumer, WriteBufferByteBased buffer) {
         int transactionId = channelTransactionManager.getTransactionIdentifier();
 
         //TODO: We need to split large messages up into chunks if it is larger than the sendBufferSize
@@ -235,7 +236,7 @@ public class SecureChannel {
                             tokenId.set(opcuaResponse.getSecureTokenId());
                             channelId.set(opcuaResponse.getSecureChannelId());
 
-                            consumer.accept(messageBuffer.toByteArray());
+                            commonPool().submit(() -> consumer.accept(messageBuffer.toByteArray()));
                         }
                     });
             } catch (Exception e) {
@@ -266,7 +267,7 @@ public class SecureChannel {
             .expectResponse(OpcuaAPU.class, REQUEST_TIMEOUT)
             .check(p -> p.getMessage() instanceof OpcuaAcknowledgeResponse)
             .unwrap(p -> (OpcuaAcknowledgeResponse) p.getMessage())
-            .handle(opcuaAcknowledgeResponse -> onConnectOpenSecureChannel(context, opcuaAcknowledgeResponse));
+            .handle(opcuaAcknowledgeResponse -> commonPool().submit(() -> onConnectOpenSecureChannel(context, opcuaAcknowledgeResponse)));
         channelTransactionManager.submit(requestConsumer, channelTransactionManager.getTransactionIdentifier());
     }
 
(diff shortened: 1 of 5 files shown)


hutcheb:
I don't completely understand the need to have the response handlers in a separate pool, but certainly not going to hold up this PR. It looks like it is failing the github actions though.

Although at the moment I can't run the rad example, on this branch or the main develop branch.
image

@chrisdutz I feel that this may be applicable to other drivers. What are your thought on it?

takraj:
@hutcheb The response handlers are executed on the Netty Event Loop thread, similarly to user events, and packetization. With this PR, I have introduced a blocking wait for packetization completion, so in case of a user event handler or a response handler tries to send a new OPC-UA message, it will deadlock, because the Event Loop Thread is still in use. Thus I moved them to seperate threads.

There might be better solutions for this, but I tried to avoid large architectural changes.

I'll check why the tests are failing. I have run them like 1000 times on my PC, with success, so I guess this must be something related to a race condition again.

hutcheb:
It may be the specific OPCUA server that is being tested against. It looks like the change in timeout from the start of the keepAlive() method (A previous PR) is causing the issue. Moving it back should fix it.

I think there is a requirement to not send the keep alive before 75% of the lifetime period, sending it straight away is the problem. There also could be an issue between the session timeout and the lifetime period but I'd have to look it up.

But this PR isn't causing that issue.

hutcheb:
image

hutcheb:
Yeah looks like the sleep should also be using the lifetime period and not the session timeout.

takraj:
@hutcheb I have made adjustments to the keepalive thread in: https://github.com/apache/plc4x/pull/1148

Please check.

takraj:
All tests are failing at `PlcConnection.close()`, with this exception:

```
java.util.concurrent.RejectedExecutionException: event executor terminated
	at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:934)
	at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:351)
	at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:344)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:836)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute0(SingleThreadEventExecutor.java:827)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:817)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:378)
	at io.netty.channel.DefaultChannelPipeline.fireUserEventTriggered(DefaultChannelPipeline.java:913)
	at org.apache.plc4x.java.spi.connection.DefaultNettyPlcConnection.close(DefaultNettyPlcConnection.java:174)
	at org.apache.plc4x.java.opcua.OpcuaPlcDriverTest$ConnectionRelated.connectionWithPlcAuthentication(OpcuaPlcDriverTest.java:292)
```

I think I was able to track down the problem, and I have a fix proposal: https://github.com/apache/plc4x/pull/1151

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/context/SecureChannel.java
```diff
@@ -361,16 +362,17 @@ public void onConnectOpenSecureChannel(ConversationContext<OpcuaAPU> context, Op
                             LOGGER.error("Failed to connect to opc ua server for the following reason:- {}, {}", ((ResponseHeader) fault.getResponseHeader()).getServiceResult().getStatusCode(), OpcuaStatusCode.enumForValue(((ResponseHeader) fault.getResponseHeader()).getServiceResult().getStatusCode()));
                         } else {
                             LOGGER.debug("Got Secure Response Connection Response");
-                            try {
-                                OpenSecureChannelResponse openSecureChannelResponse = (OpenSecureChannelResponse) message.getBody();
-                                ChannelSecurityToken securityToken = (ChannelSecurityToken) openSecureChannelResponse.getSecurityToken();
-                                tokenId.set((int) securityToken.getTokenId());
-                                channelId.set((int) securityToken.getChannelId());
-                                lifetime = securityToken.getRevisedLifetime();
```
takraj, line 369: It seems like `git merge` has somehow removed this line... I'll add one more commit to fix this.

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/context/SecureChannel.java
```diff
@@ -361,16 +362,17 @@ public void onConnectOpenSecureChannel(ConversationContext<OpcuaAPU> context, Op
...
-                                onConnectCreateSessionRequest(context);
-                            } catch (PlcConnectionException e) {
-                                LOGGER.error("Error occurred while connecting to OPC UA server", e);
-                            }
+                            OpenSecureChannelResponse openSecureChannelResponse = (OpenSecureChannelResponse) message.getBody();
+                            ChannelSecurityToken securityToken = (ChannelSecurityToken) openSecureChannelResponse.getSecurityToken();
+                            tokenId.set((int) securityToken.getTokenId());
+                            channelId.set((int) securityToken.getChannelId());
```
takraj, line 368:
```suggestion
                            channelId.set((int) securityToken.getChannelId());
                            lifetime = securityToken.getRevisedLifetime();
```

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/context/SecureChannel.java
takraj: Done in 434acb5

hutcheb:
@takraj, sorry that was probably my fault. I tried to resolve an issue with he main branch the other day. Thanks again for the PR

PR #1148, opened 2023-10-14
plc4j-driver-opcua: Move back to using session lifetime * 75% as keepalive period

takraj:
As per discussion on PR: https://github.com/apache/plc4x/pull/1147


hutcheb:
Thanks again @takraj, can confirm that to now be working.

PR #1151, opened 2023-10-15
Fix race condition causing `RejectedExecutionException` on `PlcConnection.close()`

takraj:
The remote end might actively close the connection as a response to disconnect event, which can cause the Netty channel to be already closed when we want to explicitly close it on our side.

Can be easily tested in OPC-UA driver, with a few seconds delay added before the original code block:

```java
Thread.sleep(10000); // <-- delay

channel.pipeline().fireUserEventTriggered(new CloseConnectionEvent());
channel.close().awaitUninterruptibly();
```


hutcheb:
I can't think of any negatives with this PR, it seems to handle the case mentioned better than what's there already.  Thanks @takraj

Issue #1144, opened 2023-10-11
[Bug]: plc4j-driver-opcua   OPC UA Connection Address Processing Exception


qtvbwfn:
### What happened?

When no transportEndpoint is configured, the URL is not handled correctly.

UAExpert works ok
![image](https://github.com/apache/plc4x/assets/5848026/16ff127f-3019-496f-9134-d0659294c31d)

plc4j  choose endpointurl error

String connectionString = "opcua:tcp://192.168.199.100:4840?discovery=false";

![image](https://github.com/apache/plc4x/assets/5848026/60527d6d-5300-4b17-875e-30fd31155cfb)

The correct URL should be opc.tcp://192.168.199.100:4840/freeopcua/server/.

### Version

0.10.0

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [X] OPC-UA
- [ ] S7

PR #1157, opened 2023-10-18
OPC UA priority judgment using discovery parameter

qtvbwfn:
Fixed issue where short URLs without a transportEndpoint cannot establish a connection when using the discovery=false parameter

bug:
https://github.com/apache/plc4x/issues/1144


sruehl (approved):
LGTM

PR #1159, opened 2023-10-18
plc4j-driver-opcua: Re-enable a disabled test

takraj:
After running it 600 times in a row with success, it does not seem to be flaky anymore.

### DIFF
--- a/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
+++ b/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
@@ -473,10 +473,9 @@ public class OpcuaPlcDriverTest {
     }
 
     /*
-        Test added to test the syncronized Trnasactionhandler.
-        The test originally failed one out of every 5 or so.
+        Test added to test the synchronized TransactionHandler.
      */
-    //@Test
+    @Test
     public void multipleThreads() throws Exception {
         class ReadWorker extends Thread {
             private final PlcConnection connection;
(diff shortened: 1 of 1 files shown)


sruehl:
Yeah, I guess we can enable it then again. However there was something going on on different platforms. If @chrisdutz doesn't have any objection I think it is fine to merge

sruehl:
@takraj oh and just a minor nitpick: Try to write your commits like `test(plc4j/opcua): bla bla` (https://www.conventionalcommits.org/en/v1.0.0/). I usually edit them before merge but this way it can't be forgotten :)

chrisdutz:
I think some of the tests in OPC-UA were not flaky on all platforms ... there is a number of tests, that simply don't run on Docker or Paralells VMs (No idea why) ... so if we comment in this test and it stats causing problems, we just add the comments to exclude them on Docker or Paralells.

chrisdutz:
Mabe in general it would be cool, when re-enabling OPC-UA tests, to add a comment that they were re-enabled and if they were causing problems, that first Docker and Paralells should be checked ... this way I don't have to track down the problem for the 10th time ;)

plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
```diff
@@ -473,10 +473,9 @@ public void writeVariables() throws Exception {
     }
 
     /*
-        Test added to test the syncronized Trnasactionhandler.
-        The test originally failed one out of every 5 or so.
+        Test added to test the synchronized TransactionHandler.
```
sruehl:
```suggestion
        Test added to test the synchronized TransactionHandler. (This was disabled before being enabled again so it might be a candidate for those tests not running properly on different platforms)
```

takraj:
> @takraj oh and just a minor nitpick: Try to write your commits like `test(plc4j/opcua): bla bla` (https://www.conventionalcommits.org/en/v1.0.0/). I usually edit them before merge but this way it can't be forgotten :)

@sruehl OK, will do so. I was looking for such rules in the [Contributing](https://plc4x.apache.org/developers/contributing.html) guide, but have not found any.

sruehl:
Yeah, good point, I should add that there

PR #1163, opened 2023-10-18
feat(plc4j/spi): Add option to synchronously await response from PLC

takraj:
There might be cases when the driver needs to wait until the device responds our last request, before sending the next request. This feature attempts to make this more convenient.

### DIFF
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
@@ -19,6 +19,7 @@
 package org.apache.plc4x.java.spi;
 
 import io.netty.channel.Channel;
+import java.util.concurrent.ExecutionException;
 import org.apache.plc4x.java.api.authentication.PlcAuthentication;
 import org.apache.plc4x.java.api.exceptions.PlcRuntimeException;
 import org.apache.plc4x.java.spi.configuration.Configuration;
@@ -91,6 +92,8 @@ public interface ConversationContext<T> {
         boolean isDone();
 
         void cancel();
+
+        void awaitResponse() throws InterruptedException, ExecutionException;
 
     }
 
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
@@ -152,6 +152,11 @@ public class Plc4xNettyWrapper<T> extends MessageToMessageCodec<T, Object> {
                 continue;
             }
             // Timeout?
+            if (registration.isDone()) {
+                logger.debug("Removing {} as it's already done. Timed out?", registration);
+                iter.remove();
+                continue;
+            }
             logger.trace("Checking handler {} for Object of type {}", registration, t.getClass().getSimpleName());
             if (registration.getExpectClazz().isInstance(t)) {
                 logger.trace("Handler {} has right expected type {}, checking condition", registration, registration.getExpectClazz().getSimpleName());
@@ -231,6 +236,9 @@ public class Plc4xNettyWrapper<T> extends MessageToMessageCodec<T, Object> {
             completionCallback.andThen(handler.getPacketConsumer()),
             handler.getOnTimeoutConsumer(),
             handler.getErrorConsumer(),
+            handler::confirmHandled,
+            handler::confirmError,
+            handler::cancel,
             handler.getTimeout()
         );
         deferred.set(registration);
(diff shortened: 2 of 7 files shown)


chrisdutz (approved):
Checked it out and built with all modules and all tests ... no problems ;)

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
```diff
@@ -92,6 +93,8 @@ interface ContextHandler {
 
         void cancel();
 
+        void awaitResponse() throws InterruptedException, ExecutionException;
```
splatch, line 96: @takraj I am asking myself if there is a point in making `ContextHandler` a separate thing, since it gets more and more similar to regular `Future`. I suppose you need `awaitResponse` in your future patches, wouldn't future or completable future fit better there?

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -152,6 +152,11 @@ protected void decode(ChannelHandlerContext channelHandlerContext, T t, List<Obj
                 continue;
             }
             // Timeout?
+            if (registration.isDone()) {
```
splatch, line 155: There is earlier condition which indicates that registration is cancelled. Maybe making a switch to `isActive` or similar would let us merge both? From `Plc4xNettyWrapper` perspective it doesn't matter if handler is cancelled or timed out, its relevant if its still needed to be looped over. With changes you did I think we got cancel and done callbacks within registration where appropriate log statements can be made.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -231,19 +236,21 @@ public Duration getTimeout() {
             completionCallback.andThen(handler.getPacketConsumer()),
             handler.getOnTimeoutConsumer(),
             handler.getErrorConsumer(),
+            handler::confirmHandled,
+            handler::confirmError,
+            handler::cancel,
```
splatch, line 241: I'm trying to grasp with above three lines, cause I believe they are needed, but I would like to fully understand why they are added. Are we missing error callback on driver side? Anyhow, would be good if you could leave a comment in this place. If you spotted a bug then it would be great to have a test which will prevent us from getting regressions.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -231,19 +236,21 @@ public Duration getTimeout() {
...
 
     private Consumer<TimeoutException> onTimeout(AtomicReference<HandlerRegistration> reference, Consumer<TimeoutException> onTimeoutConsumer) {
-        return new Consumer<TimeoutException>() {
-            @Override
-            public void accept(TimeoutException e) {
-                registeredHandlers.remove(reference.get());
-                onTimeoutConsumer.accept(e);
-            }
+        return timeoutException -> {
+            final HandlerRegistration registration = reference.get();
```
splatch, line 250: Just a note, syntax sugar makes stack traces harder to follow. Explicit creation of `new Consumer` was intentional to keep stacktrace more obvious.

splatch (approved):
I've left some minor comments, I would appreciate message back on these, however even with these this PR looks good to go.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
takraj: OK, I wasn't aware of that. I'll change it back, and leave a comment.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
takraj: Yea, it's pretty much like a regular `Future`. The implementor `DefaultContextHandler` is basically only a proxy of another `Future` (i.e. `HandlerRegistration`), but seems to also trigger a callback, when cancelled. I'm not sure if we can get rid of that... But still, it could simply implement the normal `Future<Void>` interface instead of this one, and could be renamed to something like `CancellationObserverFuture` (?).

Did you think of something like this?

It would be even better if this `Future` could contain the response, and would `completeExceptionally()` in case of error/timeout. So instead of specifying a callback for these events, one could wait on the future, and then continue with processing the response.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
takraj: Yes, I think you are right, these two could be merged together. Moreover, if `isCancelled()` is `true`, `isDone()` is also `true`, so it would be already enough to check the latter.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
takraj: Here, a new `HandlerRegistration` is created, copying the original, in order to inject a `completionCallback` before calling the `packetConsumer`. As `confirmHandled()` and `cancel()` are called for this new `HandlerRegistration`, the original one never became done, so `ConversationContext.isDone()` has never returned `true`.

I'll try to add test for this.

splatch: Its my bug then and test should belong to me! ;-)

PR #1165, opened 2023-10-19
chore(plc4j/bacnet): Vendor IDs update

takraj:


sruehl:
seems like the golang part is missing here. If you run a full `mvn install` it should update this too

sruehl (changes requested):
add the generated golang files too

chrisdutz:
This PR was sort of made obsolete by another commit. Thanks for creating it however. Highly appeciated.

PR #1166, opened 2023-10-19
feat(plc4j/drivers): Create maven meta package to include all drivers

takraj:
This way users do not need to explicitly list all drivers in their projects, and to update it whenever a driver is added or removed.

### DIFF
--- a/plc4j/drivers/all/pom.xml
+++ b/plc4j/drivers/all/pom.xml
@@ -29,7 +29,7 @@
     <artifactId>plc4j-driver-all</artifactId>
     <packaging>pom</packaging>
 
-    <name>PLC4J: Drivers meta package</name>
+    <name>PLC4J: All Drivers</name>
     <description>Meta package to include all standard drivers</description>
 
     <dependencies>
--- a/plc4j/drivers/all/pom.xml
+++ b/plc4j/drivers/all/pom.xml
@@ -1,0 +1,146 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+  -->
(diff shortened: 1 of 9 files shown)


plc4j/drivers/all/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+    <parent>
+        <groupId>org.apache.plc4x</groupId>
+        <artifactId>plc4j-drivers</artifactId>
+        <version>0.12.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>plc4j-driver-all</artifactId>
+    <packaging>pom</packaging>
+
+    <name>PLC4J: Drivers meta package</name>
```
splatch: Its a vise practive to keep name and artifact id close. Meta package differs enough from all to lead sometimes to confusion.

takraj: @splatch Which one would you find better to change?

splatch: I think changing name would be better, cause it will keep artifact id short. :) I don't think we will ever get a driver with "all" shortcut so its safe to stay under drivers.

takraj:
```suggestion
    <name>PLC4J: All Drivers</name>
```

splatch: Looks good :)

plc4j/examples/plc4j-s7event/pom.xml
```diff
@@ -72,6 +63,13 @@
       <version>0.12.0-SNAPSHOT</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.plc4x</groupId>
+      <artifactId>plc4j-driver-all</artifactId>
```
chrisdutz: The S7event example is strongly tied to only the S7 driver ... not sure it makes sense to use the all-drivers module here.

plc4j/tools/plc4x-server/pom.xml
```diff
@@ -88,15 +88,16 @@
 
     <dependency>
       <groupId>org.apache.plc4x</groupId>
-      <artifactId>plc4j-driver-simulated</artifactId>
+      <artifactId>plc4j-driver-plc4x</artifactId>
```
chrisdutz, line 91: Woudln't the plc4x driver be part of the all package?

takraj: Yes, but in `plc4x-server` this is the only driver that is needed at compile time too, not just runtime, because this application uses types from that driver.

plc4j/examples/plc4j-s7event/pom.xml
takraj: OK, will change it to include that one driver only.

takraj: Done in e675429

plc4j/tools/plc4x-server/pom.xml
chrisdutz: Odd ... why is this driver needed at compile-time?

chrisdutz: LOL ... forgett it ... stupid me ;) ... It#s the PLC4X server, which is the server host for the PLC4X client driver. Of cours does it require these types. I was sort of thinking of the OPC-UA Server module instead ... please ignore my mumbling ;-)

PR #1167, opened 2023-10-19
feat(plc4x-server): Build a standalone jar too

takraj:
Build a fat jar that contains all the dependencies, so user can set up a plc4x proxy server much easier.

```
$ java -jar plc4j-plc4x-server-0.12.0-SNAPSHOT-uber-jar.jar
11:50:07.822 [main] DEBUG io.netty.util.internal.logging.InternalLoggerFactory -- Using SLF4J as the default logging framework
11:50:07.826 [main] DEBUG io.netty.channel.MultithreadEventLoopGroup -- -Dio.netty.eventLoopThreads: 24
11:50:07.833 [main] DEBUG io.netty.util.concurrent.GlobalEventExecutor -- -Dio.netty.globalEventExecutor.quietPeriodSeconds: 1
11:50:07.839 [main] DEBUG io.netty.util.internal.InternalThreadLocalMap -- -Dio.netty.threadLocalMap.stringBuilder.initialSize: 1024
11:50:07.839 [main] DEBUG io.netty.util.internal.InternalThreadLocalMap -- -Dio.netty.threadLocalMap.stringBuilder.maxSize: 4096
...
```

### DIFF
--- a/plc4j/tools/plc4x-server/pom.xml
+++ b/plc4j/tools/plc4x-server/pom.xml
@@ -54,6 +54,49 @@
           </execution>
         </executions>
       </plugin>
+      <!-- Build a fat jar containing all dependencies -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>generate-uber-jar</id>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <transformers combine.children="append">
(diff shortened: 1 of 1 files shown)


sruehl:
LGTM, @chrisdutz any objections?

chrisdutz:
No objections to building fat jars, however I would not want to use these as maven dependencies in the project.

Also do I think that we don't need to explicitly attach the uber-jar as I think I remember the shade plugin does that for us.

chrisdutz:
Just noticed that you're only talking about the server ... sure ... that's not used from anywhere else.

PR #1172, opened 2023-10-23
feat: Moved the code-generation into a separate profile

chrisdutz:
Added a new maven profile `update-generated-code` to the build, that allows explicitly enabling code-generation.

This is needed in preparation to supporting reproducible-builds.


chrisdutz:
In theory we could simply run the dynamic mspec-generation in a separate profile, but also only updating the real code generation seems to have made more sense.

sruehl (approved):

PR #1202, opened 2023-11-12
upgrade Felix maven-bundle-plugin

hboutemy:
fixes a vast majority (56) of Reproducible Builds issues found in previous releases (up to 1.11.0)
see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/plc4x/plc4x/README.md

there are other issues, but this one is causing much easy to fix noise: next release will be easier to review once this one is fixed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1086,7 +1086,7 @@
         <plugin>
           <groupId>org.apache.felix</groupId>
           <artifactId>maven-bundle-plugin</artifactId>
-          <version>5.1.8</version>
+          <version>5.1.9</version>
         </plugin>
 
         <plugin>


chrisdutz:
Thanks for this ... hopefully with 0.11.0 we'll finally reach something reproducible :-)

hboutemy:
I'm currently trying to calculate the diff of 0.11.0 ignoring the bundle issues, to be able to anticipate remaining issues: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/plc4x/plc4x/plc4x-parent-0.11.0.diffoscope

if you know how `protocols/bacnetip/bacnet-vendorids.mspec` is created, this is the second source of non-reproducible bits

sruehl:
this file is created the same way as the ones for knx-vendors for example.

sruehl:
@chrisdutz didn't you move all this to profiles?

chrisdutz:
Yeah ... I thought I had recently moved all of these into separate profiles: "update-generated-code", so in 0.11.0 this was still an issue, but in the current dev-version it should no longer be.

PR #1214, opened 2023-11-18
Issue/s7h

glcj:
1. Fixes problems with the pipes created by the SPI and the driver architecture.
2. URL handling is corrected.
3. Reading example S71200, S7400H and Cyclic Subscription for S7300 and S7400 are improved.


chrisdutz (commented):
It seems you moved the code generation or of the code generation profile. I please don't immediately merge this, I think it needs some cleaning up.

chrisdutz:
Ok... Just saw that it's already merged 😔 guess I'll be doing some cleaning up today then.

PR #1361, opened 2024-01-24
fix: added timeouts for reading data/getting connection

geneross:


chrisdutz:
LGTM

chrisdutz:
However a bit more explanations on what a PR is about would have been nice ;-)

geneross:
Thanks for paying attention. I was desperate to get an answer.  Lost in permissions etc. 
In general we've faced with case when `TriggeredScraperImpl `threads just silently dies partially because of network malfunction partially because of server bugs. 
So `TriggeredScraperImpl` got 2 debug threads that watch on working threads and it static get connection method now re-throw unchecked `PlcRuntimeException` because otherwise it might re-throw checked exception that leads to working thread termination. 
Second. `PlcReadRequest.builder` from `LeasedPlcConnection` now return class with execution timeout for innerPlcReadRequest.
Third. `ConnectionContainer` set container value to null when it can't get connection from the manager while trying to invalidate cached connection. Otherwise it created new `LeasedConnection` with same broken connection.

My apologies for brevity, I have to recall what I did 2 month ago. ))

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -120,14 +126,18 @@ public PlcReadRequest build() {
                 return new PlcReadRequest(){
                     @Override
                     public CompletableFuture<? extends PlcReadResponse> execute() {
-                        CompletableFuture<? extends PlcReadResponse> future = innerPlcReadRequest.execute();
+                        CompletableFuture<? extends PlcReadResponse> future =
+                            innerPlcReadRequest.execute().orTimeout(Math.min(1000,maxUseDuration.toMillis()), TimeUnit.MILLISECONDS);
```
chrisdutz, line 130: So as far as I remember the ConnectionPool ... that used to have a timer to invalidate a connection if being used too long ... but I guess in that case possibly the read operation never timed out, right? So this is an additional abort on the side of the user, correct?

chrisdutz (approved):

PR #1419, opened 2024-02-22
Site integration 3.x.x Migration - Doxia Modules 3+

ottlukas:
Rename the maven-site-plugin dependency from asciidoctor-maven-plugin to asciidoctor-converter-doxia-module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1314,7 +1314,7 @@
             <!-- Add support for asciidoctor -->
             <dependency>
               <groupId>org.asciidoctor</groupId>
-              <artifactId>asciidoctor-maven-plugin</artifactId>
+              <artifactId>asciidoctor-converter-doxia-module</artifactId>
               <version>${asciidoctor.maven.plugin.version}</version>
             </dependency>
             <dependency>


ottlukas:
https://docs.asciidoctor.org/maven-tools/latest/site-integration/v3-migration-guide/ ->Rename the maven-site-plugin dependency from asciidoctor-maven-plugin to asciidoctor-converter-doxia-module.

chrisdutz:
The asciidoctor folks have an open PR for this, but they said they will only work on this as soon as the site plugin is final ... so I guess we have to wait for the site plugin to be finally released and then for the asciidoctor folks to update the plugin. Till then we should stay at 2.x.x

ottlukas:
> The asciidoctor folks have an open PR for this, but they said they will only work on this as soon as the site plugin is final ... so I guess we have to wait for the site plugin to be finally released and then for the asciidoctor folks to update the plugin. Till then we should stay at 2.x.x

@chrisdutz can you link in the mentioned Pull request in asciidoctor? Thank you!

chrisdutz:
I have been looking into updating the site and the asciidoctor plugin ... however this seems to require quite some work and possibly a new site skin.

chrisdutz:
Here's the PR in the asciidoctor-maven-plugin that I was talking about: https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/685

hutcheb:
This shouldn’t be relevant anymore due to the website update but can take a better look on the weekend.

PR #1446, opened 2024-03-11
CANopen driver fixes

splatch:
Small updates needed to get driver working again after changes in 0.11/0.12 releases.


splatch:
@chrisdutz I've switched it from draft as it runs stable with my machine, could you give a spin to it? These are minor change but I enable some of new tests which should not break under mac/windows etc..

chrisdutz (approved):
The only changes I really care about are the ones in SPI and it seems your renaming actually made it comply with the naming convention in API ... so I'm fine with all of it.

splatch:
Great, I'll try to push one more thing related to tag handling for generic can and then, if CI/CD is happy, merge it.

PR #1453, opened 2024-03-18
Update PlcRawByteArray.java

mrwhy-orig:
Removed method duplicate


splatch:
Yeah, looks like its a result of failed merge.

PR #1468, opened 2024-03-26
Convert the string address into a PlcTag

qtvbwfn:
Convert the string address into a PlcTag. 
For string tagAddress verification


chrisdutz:
I would prefer to call the method differently ... 

parseTagAddress

And not to have the "null" variant.

plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcReadRequest.java
```diff
@@ -38,6 +39,16 @@ interface Builder extends PlcRequestBuilder {
...
+        /**
+         * Parses a string representation of a tag address into a {@link PlcTag} object.
+         * <p>
+         * This method safely attempts to convert the given tag address string into a PlcTag, returning an
+         * {@link Optional} that is empty if the parsing fails.
+         *
+         * @param tagAddress The string representation of the tag address to parse.
+         * @return An optional holding the parsed PLC tag if successful, otherwise an empty optional.
+         */
+        Optional<PlcTag> parseSafe(String tagAddress);
```
chrisdutz: I would prefer "parseTagAddress" for this to match the naming pattern of the other methods.

chrisdutz: Also not 100% sure, if we should not instead have this method in the connection instead ... reasoning is that when adding it to the builder it doesn't really add anything to the request it builds .... however having a "parseTagAddress" and "parseQueryAddress" in the connection sounds like a very cool thing to do, as this way tooling utilizing PLC4X can use to validate tags and queries.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/messages/DefaultPlcReadRequest.java
```diff
@@ -139,6 +136,19 @@ public PlcReadRequest build() {
             return new DefaultPlcReadRequest(reader, parsedTags);
         }
 
+        public PlcTag parseOrNull(String tagAddress){
```
chrisdutz: Generally the same comment as for the previous, in addition I personally don't really like the "or-null" variant ... I would prefer to keep only the optional one.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -176,6 +176,11 @@ public PlcReadRequest.Builder addTagAddress(String name, String tagAddress) {
             public PlcReadRequest.Builder addTag(String name, PlcTag tag) {
                 return innerBuilder.addTag(name, tag);
             }
+
+            @Override
+            public Optional<PlcTag> parseSafe(String tagAddress) {
```
chrisdutz: Same comment as, I think this would be a better fit in the connection itself, not in the builder.

chrisdutz (changes requested):
Hi,

so I had a more detailed look at the PR and I generally like the idea of having a method to parse a tag-address into a tag as this can easily be used by tools to validate and verify the addresses users entered. 

As mentioned before, I would probably prefer a name: parseTagAddress and to only have the variant to return an optional.

I would probably also add a parseQueryAddress.

However, I don't think adding them to the builders is the ideal way to do it. I would much more prefer to have them added to the connection object directly, as the result of the method doesn't change the state of the Builder.

qtvbwfn:
The method name has been changed from `parseSafe` to `parseTagAddressSafe` for better readability.

chrisdutz:
Hi,

thank you for adressing part of my comments, however we use Optionals in quite a number of places, so adding the "safe" suffix would not ex expected here. So in general I would propose a name without the "Safe" ending.

But the more important thing is, that I think the parseTagAddress methods should probably not be located in the builder, but in the connection ... as I mentioned before, the parseTagAddress method does not change the internal state of the builder, so I guess this is not the right place to it to be located.

However your PR did bring up the general question of adding a tagAddress parser method and also start me thinking about the general use of our queryAddress concept ... on the mailinglis I've initated a discussion on using only TagAddress objects and having special Query Tag types for protocols that support that.

Chris

qtvbwfn:
Perhaps it is because I am not very familiar with the code structure. Based on the Java examples provided on the official website, one typically starts by obtaining the connection object `PlcConnection`. However, it does not seem straightforward to add methods for parse the string address . On the commonly used `PlcReadRequest.Builder`, there is a `tagHandler` object that can be utilized to manage this matter. I am not sure how to describe it precisely, but it is merely for the convenience of use.

The method name has been changed from `parseTagAddressSafe` to `parseTagAddress` for better readability.

     Thank you for your reply and the associated comments.     @chrisdutz

chrisdutz:
Well you can get TagObjects from the result of a Browse request ... that was why we added the addTag in paralall to addTagAddress methods. 

Let's see what the other members of the project think ... I'll ping them to reply here. Might be that they see it the same way you do it. So let's ask them.

I guess the main use-case that I was thinking of, was that adding it to the connection, would be a general purpose tool that allows other tools to also simply validate tag-addresses.

splatch:
Moving proposed method to connection fits better, because there are drivers for which we either have write (publish) or subscribe - for example Generic CAN. From future point of view, moving it to connection will also fit listen only radio/broadcast protocols, where we have no way to explicitly read information.

qtvbwfn:
The code position has been adjusted to use the method of overriding the `parseTagAddress` of `PlcConnection` for string address validation.

@chrisdutz

plc4j/drivers/ctrlx/src/main/java/org/apache/plc4x/java/ctrlx/readwrite/connection/CtrlXConnection.java
```diff
@@ -90,6 +93,17 @@ public void close() throws Exception {
         executorService.shutdown();
     }
 
+    @Override
+    public Optional<PlcTag> parseTagAddress(String tagAddress) {
+        PlcTag plcTag;
+        try {
+            plcTag = controlXTagHandler.parseTag(tagAddress);
+        } catch (Exception e) {
```
sruehl, line 105: the Exception should be at least logged at a debug level or if it never matters the variable should be renamed to `ignore`

plc4j/drivers/mock/src/main/java/org/apache/plc4x/java/mock/connection/MockConnection.java
```diff
@@ -221,4 +224,14 @@ public PlcAuthentication getAuthentication() {
         return authentication;
     }
 
+    @Override
+    public Optional<PlcTag> parseTagAddress(String tagAddress) {
+        PlcTag plcTag;
+        try{
+            plcTag = mockTagHandler.parseTag(tagAddress);
+        }catch (Exception e){
```
sruehl: the Exception should be at least logged at a debug level or if it never matters the variable should be renamed to `ignore`

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/connection/AbstractPlcConnection.java
```diff
@@ -228,4 +230,14 @@ public CompletableFuture<PlcBrowseResponse> browseWithInterceptor(PlcBrowseReque
         return protocol.browseWithInterceptor(browseRequest, interceptor);
     }
 
+    @Override
+    public Optional<PlcTag> parseTagAddress(String tagAddress) {
+        PlcTag plcTag;
+        try{
+            plcTag = tagHandler.parseTag(tagAddress);
+        }catch (Exception e){
```
sruehl: the Exception should be at least logged at a debug level or if it never matters the variable should be renamed to `ignore`

plc4j/drivers/ctrlx/src/main/java/org/apache/plc4x/java/ctrlx/readwrite/connection/CtrlXConnection.java
chrisdutz: This is no real driver yet ... it's a skeleton, which I started working on, but then got all time stolen to continue working on it :-/

qtvbwfn:
Thank you for your attention to the RP I submitted. The existence of a verification method during use is indeed beneficial for the subsequent data reading work. I appreciate your contributions to the project.

sruehl (approved):

chrisdutz:
Hmmm ... so we now have a parseTag on Driver and Connection ... I'm thinking it might be better to give the driver-version a slightly different name. The parse method there can technically only parse on a much higher level. The version on the connection-level knows the type of device and can offer a more-correct answer. But generally this PR looks good to me.

chrisdutz (approved):

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
```diff
@@ -80,6 +80,11 @@ public synchronized void close() {
         connectionContainer.returnConnection(this, invalidateConnection);
     }
 
+    @Override
+    public Optional<PlcTag> parseTagAddress(String tagAddress) {
+        return Optional.empty();
```
chrisdutz: I think this should probably forward to the driver being used and not simply return empty.

chrisdutz (changes requested):
Sorry ... just found something we might want to change.

plc4j/tools/connection-cache/src/main/java/org/apache/plc4x/java/utils/cache/LeasedPlcConnection.java
qtvbwfn: > I think this should probably forward to the driver being used and not simply return empty.

Requesting assistance, as I am not very familiar with the overall code of the project, PlcTagHandler was not found within LeasedPlcConnection. However, LeasedPlcConnection is an implementation class of PlcConnection. Is it appropriate to directly use connection.get().parseTagAddress(tagAddress)?

chrisdutz: Leased connection has a "connection" property which refers to the actual connection ... so I guess something like this should work:

```
    @Override
    public Optional<PlcTag> parseTagAddress(String tagAddress) {
        PlcConnection plcConnection = connection.get();
        if(plcConnection == null) {
            throw new PlcRuntimeException("Error using leased connection after returning it to the cache.");
        }
        return plcConnection.parseTagAddress(tagAddress);
    }

```

qtvbwfn: Already modified.

chrisdutz (approved):

Issue #1509, opened 2024-04-11
[Bug]: [ADS] Writing a List of Struct writes nothing


IsmoLeszczynski:
### What happened?

The PLC I'm developing against has a symbol of type List that contains a list of Structs, which aren't exposed as their own symbols (at least not included in the browse results when retrieving "symbols"). Reading the current value from the symbol produces the correct List of Struct results, but when I try to write an updated value back to it, the driver resolves it to an empty byte[]. This seems to be happening within `AdsProtocolLogic.serializeInternal()`, when iterating the list entries it determines each entry to be a list as well, since the same dataType is used for the child iteration as well, and the `DataItem.staticSerialize()` can't handle that.

How the tag is seen in TwinCAT:
![TwinCAT_screenshot](https://github.com/apache/plc4x/assets/156197448/f35b44f5-53b3-454f-8005-b418524c90dc)

The browsed symbol:
![TestTag](https://github.com/apache/plc4x/assets/156197448/632c3ed8-dec7-4457-ace9-caf1cd347656)

The attempted write value:
![TestValue_ListOfStruct](https://github.com/apache/plc4x/assets/156197448/37c2442d-b775-4335-9732-43008a693f43)

### Version

v0.12.0

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [X] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] S7

chrisdutz:
Could you add wireshark recordings of this? I only have 4h of work on PLC4X per week and I would want to make it as efficient as possible.

IsmoLeszczynski:
Sure, here you go:
[ads-invalid-write-list-of-struct.zip](https://github.com/apache/plc4x/files/14956297/ads-invalid-write-list-of-struct.zip)

The capture contains the connection, browsing of symbols, reading the List-of-Struct symbol and writing back to it - the data bit in the write is completely empty.

I also created a PR(#1517) that fixed my use case, the issue seemed to only be that for Lists the children were also processed as Lists.

IsmoLeszczynski:
And I apparently closed this by mistake, my bad.

[...]

PR #1517, opened 2024-04-12
fix(plc4j/ads): Fixed List serialization passing invalid datatype

IsmoLeszczynski:
Fix for: #1509 

The List serialization for ADS passed the main List type to each child element when serializing them instead of determining the correct type for the children. This fix modifies the List iteration to determine the simple type from the List type, which is the actual type for each child element.

Regarding the `arrayInfo.subList()`, I'm not familiar enough to say if that should also be determined from the child type.


chrisdutz:
Having a deeper look ... I am not sure this is a universal fix for the problem. In 90% of the entries in the dataTypeTable the simpleDataTypeName is "". I think it's only used in cased of refereces, Arrays and some of the system types. I would assume you're reading one of these system types, so I guess your fix would make that work, but break the regular DTOs. 

I'll have a deeper look into to the problem.

chrisdutz:
Zooming a bit out of the code ... you are absolutely correct ... this part of the code handles the array serialization. And in this case the datatype will always be an "Array of LALALA" and in that case simple type will always contain "LALALA" ...
Thanks for spotting this ;-)

chrisdutz (approved):

Issue #1433, opened 2024-02-27
[Bug]: ads driver, writing multiple tags in 1 writeRequestBuilder doesnt work anymore


patrickboisclair:
### What happened?

Hi all,

Beckhoff ads driver.

When trying to write multiple TAGS at once, seems to be broken at the moment. 
Doing the same thing with 0.9.1 works just fine.

Example:

```java
            PlcWriteRequest writeRequest = plcConnection.writeRequestBuilder()
                                                        .addTagAddress("ValX", "MAIN.ValX", new PlcDINT(154))
                                                        .addTagAddress("ValY", "MAIN.ValY", new PlcDINT(178))
                                                        .build();

            PlcWriteResponse writeResponse = writeRequest.execute().get(); 
``` 
Results in:

Exception in thread "main" java.util.concurrent.ExecutionException: org.apache.plc4x.java.api.exceptions.PlcException: Error
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
        at com.epsilia.Beckhoff.execute(Beckhoff.java:48)
        at com.epsilia.Main.main(Main.java:6)
Caused by: org.apache.plc4x.java.api.exceptions.PlcException: Error
        at org.apache.plc4x.java.ads.protocol.AdsProtocolLogic.lambda$84(AdsProtocolLogic.java:1108)
        at java.base/java.util.function.Consumer.lambda$andThen$0(Consumer.java:65)
        at org.apache.plc4x.java.spi.Plc4xNettyWrapper.decode(Plc4xNettyWrapper.java:183)
        at io.netty.handler.codec.MessageToMessageCodec$2.decode(MessageToMessageCodec.java:81)
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:88)
        at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:111)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
[...]

PR #1524, opened 2024-04-12
fix(plc4j/ads): Invalid size writing multiple tags

IsmoLeszczynski:
This fixes the issue mentioned in #1433.

When writing multiple tags via ADS, the IndexOffset and tag specific data type sizes were incorrect; the IndexOffset should indicate how many tags are being written instead of how many bytes, and the individual tag write sizes were pointing to the length of the data type table entry instead of the data type size from that entry. With these changes I was able to write multiple values without issues.


chrisdutz:
Thanks for spotting this ... I guess it would have taken me a bit to find that issue. Very much appreciated ;-)
With your fix indeed my ManualAdsTest worked flawlessly (which now also does multi-item writes as part of the test)

PR #1581, opened 2024-05-03
Refactor / improve python modbus

ottlukas:


plc4py/plc4py/drivers/modbus/ModbusConfiguration.py
hutcheb, line 1: Wait, we can do that?

plc4py/plc4py/drivers/modbus/ModbusConnection.py
hutcheb, line 1: This removes the timeout. The timeout can be handled better, but I like having them in a certain points as a catch all in case it gets into the situation where we get stuck in a loop.

hutcheb (approved):

ottlukas:
This is a **draft** and experiment space for improving PLC4Py.

plc4py/plc4py/drivers/modbus/ModbusConfiguration.py
ottlukas: 1. if statement with assignment:

        This is a more explicit approach:
        
        It checks if self.transport is None using the is operator.
        If it is None, it assigns the string "tcp" to the self.transport attribute.

2. Conditional assignment (or operator):

        This is a more concise way to achieve the same result using the or operator:
        
        The expression self.transport or "tcp" evaluates to:
        self.transport if it has a value other than None.
        "tcp" if self.transport is currently None.
        The result of the expression is then assigned to self.transport.

Choosing between them:

If clarity is a priority and you want to make the logic explicit, use the if statement approach.
If code brevity is preferred, the conditional assignment with the or operator is a good choice.
Both methods are valid, and the best choice depends on your coding style and the specific context of your code.

The question of what code style we like to pursue.
Probably beeing more explicit is the way forward.

hutcheb: Wasn't complaining, I like it.

ottlukas (commented):
Working branch so far

PR #1632, opened 2024-06-03
add modbus support BIG_ENDIAN_WORD_SWAP and  LITTLE_ENDIAN_WORD_SWAP

StrawberryBlue:
modbus support BIG_ENDIAN_WORD_SWAP
modbus support LITTLE_ENDIAN_WORD_SWAP


chrisdutz:
Adding the missing header after merging.

chrisdutz:
Ok ... so it turns out that this PR had huge side-effects on protocols such as ADS, as it didn't consider uint 32 or not full short values. I had to revert the changes ... I moved the ByteOrder Helper into the modbus protocol and created a new type: ModbusByteOrder. I don't think this Word-Swap functionality is needed anywhere else than Modbus, so I would like to keep the changes there.

PR #1633, opened 2024-06-03
Setting third parameter to emptyCollection in static of methods call to SymbolicAdsTag

mrwhy-orig:
This fixes a bug in subscription where the SymbolicAdsTag gets casted to DefaultSubscriptionTag. 
It was possible that the arrayInfo field wasn't initialized which lead to a NullpointerException.


mrwhy-orig:
Maybe it is more reasonable to return a immutable empty collection in getArrayInfo method if the field is null... (If someone set it to null it is unlikely that there will be more than one value, right?)

PR #1703, opened 2024-07-16
Introduce maven dependency management section for plc4x project.

splatch:
An experimental move of versions to a single place. This allows importing of plc4x parent as BOM for downstream projects.


chrisdutz:
But as all is the same version anyway.... Not quite sure what the benefit is.

I understand in spring and spring-boot this can be challenging, as everything is in different versions.
I particularly like that you instantly know what dependency is an inner-project dependency (the ones with an explicit version) and which ones are not (the ones without).... And for the parent relationship you can't get rid of the versions anyway.

chrisdutz:
I could imagine that having a separate "bom" artifact, that defines the versions of all modules so users could simply import that as potentially being valuable, but even in that case would I leave the versions of reactor modules in the pom.

I've seen this over the last 20 years ... if you're continuously switching versions you always have to scroll up to see in the parent pom version to know where you're currently located. 

The maintenance work doesn't really exist at all as the release plugin nicely updates all of this for you.

PR #1704, opened 2024-07-17
Odot

glcj:
This modification allows the use of the TCP/MPI gateway unit from the Odot company.

The "COTP - TPDU number" field must always be zero.

This was verified by comparing commercial and open source drivers.

The Read/Write was verified, event management is pending.


splatch:
@glcj Are you sure it won't break regular S7 connectivity? As far I can see each operation we make on driver side includes tpdu reference.

glcj:
Hi Lukas,

It took me some time to evaluate that because of the strangeness of the case.

In Siemens communications, they do not update "TPDU Number" at the COTP link layer, it is always zero.

The TPDU if used incrementally at the S7 layer

The Siemens CPs discard this field, which is why the driver works, but in the particular case of the Odot equipment it will take it into account and expect it to always be zero.

I tested the driver against the CP 443-1 (S7-400) and everything worked without problems.

As indicated, there are other functionalities in the driver that form the message including the COTP layer (diagnosis request, subscription, etc.), which I must evaluate.

Kind regards

splatch:
Awesome, thanks for clarification!

Issue #1471, opened 2024-03-27
[Feature Request]: Support for software/hardware timestamp information in CANopen driver


splatch:
### What would you like to happen?

JavaCAN library (see pschichtel/JavaCAN#5) recently introduced support for access to hardware or software timestamp tracked by socketcan interfaces. We can integrate this information within our driver to better track timing of operations.

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [X] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] S7

chrisdutz:
Currently the PlcSubscriptionEvent should have a getTimestamp() method ... I think this should be what you're looking for.

However am I currently brainstorming a bit on how to refactor the SubscriptionAPI to use one callback per request and not per field, so this API is a bit in flux right now.

PR #1725, opened 2024-08-09
Draft of metadata API with timestamp tracking capabilities.

splatch:
Introduction of tag metadata which serves similar purpose to result set metadata for JDBC, but with constraints on key syntax. Softly related to #1471 and other drivers which are capable of providing timestamp information as a part of message exchanges.

In order to force registration of metadata keys, so they can be identified by callers, if needed, keys have non-string form.

Feature been discussed recently in this thread:
https://lists.apache.org/thread/chc1p0czm7q33n1r57zbw7h11c7wzp4l
but origins of feature come from older thread:
https://lists.apache.org/thread/r3g4y6p7kst7z5bdpmccd8538h3hl80f

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcMetadataKeys.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcMetadataKeys.java
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *   https://www.apache.org/licenses/LICENSE-2.0
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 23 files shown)


chrisdutz:
I'm on holidays this week can have a look next Friday.

chrisdutz:
It does feel a bit different from the other parts of our API ... but I guess I would simply need to play around with it a bit. I gess we can always change things in the future, if we want to give all parts the same feel.

chrisdutz (approved):
LGTM

splatch:
@chrisdutz FYI I've separated API and SPI parts, now it is a bit closer to overall paradigm we have. Still metadata is immutable and a bit constrained in manipulation to avoid 1) putting too much there 2) relying too much on it.

PR #1727, opened 2024-08-13
Improve error handling and propagate exceptions back to driver

### LINKED ISSUE #1726: [Bug]: The onError calls are always ignored
### What happened?

Any send operation which expects response can specify criteria for matching answers but also error handlers which should be called when handling logic fails. Determination when error handler should be called is not trivial. Most of protocols have fairly long `unwrap`/`check` statements which can fail at any stage. We can not assume which stage will fail, we can't layer error handlers, thus we need carefully craft logic and make sure that we don't catch too much and still propagate failures to driver and clean up handlers.
Currently it might be that packet consumers are retained until timeout, even if we know that they can't properly handle incoming data.

### Version

v0.12

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] S7

splatch:
Closes #1726.

### DIFF
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
@@ -149,7 +149,7 @@ public interface ConversationContext<T> {
 
         void cancel();
 
-        void awaitResponse() throws InterruptedException, ExecutionException;
+        void await() throws InterruptedException, ExecutionException;
 
     }
 
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
@@ -27,6 +27,7 @@ import io.vavr.control.Either;
 
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
 import org.apache.plc4x.java.api.authentication.PlcAuthentication;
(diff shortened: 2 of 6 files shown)


plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
```diff
@@ -149,7 +149,7 @@ interface ContextHandler {
 
         void cancel();
 
-        void awaitResponse() throws InterruptedException, ExecutionException;
+        void await() throws InterruptedException, ExecutionException;
```
chrisdutz, line 152: Why change this? I think the awaitResponse gave context to what you're waiting for?

chrisdutz (approved):
LGTM (Looks Good To Me)

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/ConversationContext.java
splatch: We use same context for server side/simulators. This statement is applicable to `expectRequest()` thus renaming it  to `await` makes it more generic and match various roles implied by use of API.

chrisdutz: Ups... Didn't want to unresolve.. Mobile phone UI slighty clunky. No objections

PR #1738, opened 2024-08-23
Add more tests for mspec parser and code generator.

splatch:


sruehl:
as of a668321 this doesnt even build on java :/

sruehl:
ok I think I was on the wrong revision e5d1782... nvm

sruehl:
@splatch please rebase on current main

sruehl:
@splatch careful that you leave it as read`A`Manual field otherwise we have lang collisions in Goland

sruehl:
also it seems you reverted stuff by accident with the rebase, not sure why dad got deleted :D

sruehl:
as a small background why I used non-technical terms: When you debug that stuff and you stumble over structs named struct it is getting pretty confusing. So the goal there was to avoid to use the codegen terms so it is easier to seperate content from codegen items.

splatch:
I was sure that this `A` was a typo! I will try to pull back commit from reflog and cherry pick it on top of develop.

sruehl:
no worries, just from the looks of it it looks like a failed rebase/merge :)
Afaik this should be the only change in that PR 
image

sruehl:
maybe the other one above but not sure. Think I got rid of it because of the confusion 
image
but if you added that feel free to readd

splatch:
@sruehl I've cherry picked only intended changes, should be better now.

chrisdutz:
What's the status of this?

PR #1743, opened 2024-09-02
Feature/s7 discovery

chrisdutz:
Fixed an issue with the Profinet auto-discovery that wouldn't work if I my wi-fi was disabled.

Added support for discovering S7-1200 and S7-1500 devices based on the Profinet auto-discovery mechanism, as every S7 device I had reacted to profinet auto-discovery requests and contained the vendor and device id. So I've duplicated the small fraction of the Profinet DCP protocol, that was needed to implment this and simply filter devices based on their device-id.

Currently I'm still missing some device ids for S7 300 and S7 400 (And possibly others.)


chrisdutz:
@ceos01 would be cool if you could add the device ids of your 300 and 400 (and if you know any other device ids of devices that support the s7 comm protocol)

glcj:
Hi Chris,

In Windows I have this message

06:54:47.748 [main] ERROR o.a.p.j.s.r.d.ProfinetChannel - Got an exception while processing raw socket data
org.pcap4j.core.PcapNativeException: Error occurred in pcap_compile: ethernet addresses supported only on ethernet/FDDI/token ring/802.11/ATM LANE/Fibre Channel
	at org.pcap4j.core.PcapHandle.setFilter(PcapHandle.java:482)
	at org.pcap4j.core.PcapHandle.setFilter(PcapHandle.java:508)
	at org.apache.plc4x.java.s7.readwrite.discovery.ProfinetChannel.getInterfaceHandles(ProfinetChannel.java:245)
	at org.apache.plc4x.java.s7.readwrite.discovery.ProfinetChannel.<init>(ProfinetChannel.java:70)
	at org.apache.plc4x.java.s7.readwrite.S7Driver.discoveryRequestBuilder(S7Driver.java:116)
	at org.apache.plc4x.java.s7.readwrite.ManualS7Discovery.main(ManualS7Discovery.java:27)

It must be a problem with the interface, disable WiFi, VPN. I use a USB/Ethernet interface which Wireshark recognizes without problems.

I keep trying

chrisdutz:
It sounds as you might have a network device that is not "ethernet/FDDI/token ring/802.11/ATM LANE/Fibre Channel" ... would be cool, if you could find out the characteristics of the device it's failing for ... I just realized I needed to check disabled devices ... doing raw stuff does require the developer to do a bit more than with the bubble-wrapped normal stack.

chrisdutz:
I think I'll merge this PR as soon as I get the build green again ... we can continue optimizing things on develop then.

PR #1744, opened 2024-09-04
Feature/modbus optimizer

chrisdutz:
Implemented an optimizer for Modbus in Java, that groups together individual items and loads chunks of bytes.
This needed quite a bit of updating which also had an effect on the generated code of almost every language we have.

However with this update the read speed of the ManualModbusTest increased by a factor of more than 10.


chrisdutz:
@hutcheb and @ottlukas I could use your help on the python stuff ... should not be a huge change ... in other languages, if I have an array of byte byte that it uses the PlcRawByteArray instead of a List.

ottlukas (approved):

hutcheb:
@chrisdutz  Yeah have to spend some time on fixing things when it comes to using the PlcValues, the Modbus write PR will include some fixes for it. But for now it passes the tests.

chrisdutz:
Did I break anything with my addition?

hutcheb:
No, it was just a Byte Array branch in the code gen that hadn't been used before. It was generating code that had a syntax error in it, that's why black couldn't reformat the file.

By the way we are using black to firstly check for syntax errors, as well as clean up some of the formatting in the code gen output. That way we can be lazy and don't have to get the formatting in the code gen output exactly right.

ottlukas:
> No, it was just a Byte Array branch in the code gen that hadn't been used before. It was generating code that had a syntax error in it, that's why black couldn't reformat the file.
> 
> By the way we are using black to firstly check for syntax errors, as well as clean up some of the formatting in the code gen output. That way we can be lazy and don't have to get the formatting in the code gen output exactly right.

[Black - The uncompromising code formatter](https://black.readthedocs.io/en/stable/)

chrisdutz:
> No, it was just a Byte Array branch in the code gen that hadn't been used before. It was generating code that had a syntax error in it, that's why black couldn't reformat the file.
> 
> By the way we are using black to firstly check for syntax errors, as well as clean up some of the formatting in the code gen output. That way we can be lazy and don't have to get the formatting in the code gen output exactly right.

As long as the output gives you a hint to what the problem is ... admittedly I was a bit lost and commented out black in order to get a sensible error report.

PR #1786, opened 2024-09-24
refactor: Fine tuning of the API and SPI

chrisdutz:
What started as something to allow returning and handing INVALID_ADDRESS responses for individual items (Currently one address string, that the parser can't handle is enough to fail the entire request).

This required to add a getPlcResponseCode to the PlcTagRequests and to introduce a new set of items for the TagItem and TagValueItems, which allow returning a non-OK return code.

Alongside these changes, I started cleaning things up a little.

In the end I would like to bring back together the Tags and Queries and to only have Tags for both Browse and the other APIs. They could be QueryTags that allow filtering, but in general I think it would be better to only have one type for all APIs.

I could need a bit of help with the other languages.


chrisdutz:
Right now I would first update the test-case xml files for the current state and as soon as that's done, I would try to update the other languages accordingly.

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/tag/SymbolicAdsTag.java
```diff
@@ -37,8 +37,7 @@
  */
 public class SymbolicAdsTag implements AdsTag {
 
-    // TODO: Model the end of this address to allow usage of multi-dimensional arrays.
-    private static final Pattern SYMBOLIC_ADDRESS_PATTERN = Pattern.compile("^(?<symbolicAddress>.+)");
+    private static final Pattern SYMBOLIC_ADDRESS_PATTERN = Pattern.compile("^([\\w_]+)(\"[\"\\d*]\")*(\\.(\\w+)(\"[\"\\d*]\")*)*");
```
sruehl, line 40: why did you remove the `symbolicAddress` reference?

sruehl:
@chrisdutz once you are in a happy state with this PR I can look into adding the Go-Variant for it

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/tag/SymbolicAdsTag.java
chrisdutz: I'm still working on this.... It's all still WIP.

chrisdutz:
Thanks... I just thought I'd push my changes for feedback as early as possible. No need to waste time in something the rest dislikes.

chrisdutz:
So currently it seems that my changes had no effect on C, Python and all other languages, only Go probably needs some adjusting @sruehl would be cool, if you could have a lootk.

sruehl (approved):

PR #1803, opened 2024-10-06
fix: Updated the subscription API to match the changes of the others

chrisdutz:
So I updated the subscription api to match the structure of the updated read- and write-api ... now also the Subscription API supports having individually invalid tags without failing the entire request.

Also refactored the subscription request to register a handler for the entire request instead of each tag.


plc4j/drivers/firmata/src/test/java/org/apache/plc4x/java/firmata/readwrite/FirmataManualTest.java
```diff
@@ -30,7 +30,7 @@ public static void main(String[] args) throws Exception {
 
         // Default for the Firmata Arduino sketch is, 57600 baud, 8 data-bits, no parity and one stop bit (Which is the default for the Serial transport)
         try(PlcConnection connection = PlcDriverManager.getDefault().getConnectionManager().getConnection("firmata:///dev/tty.usbmodem2114401")) {
-            PlcSubscriptionRequest subscriptionRequest = connection.subscriptionRequestBuilder().addEventTagAddress("analog1", "analog:1").addPreRegisteredConsumer("analog1", plcSubscriptionEvent -> {
+            PlcSubscriptionRequest subscriptionRequest = connection.subscriptionRequestBuilder().addEventTagAddress("analog1", "analog:1").setConsumer(plcSubscriptionEvent -> {
```
sruehl, line 33: shouldn't the consumer be set then as 3rd argument in `addEventTagAddress` following that logic? otherwise that would semantical not the same

plc4j/drivers/iec-60870/src/test/java/org/apache/plc4x/java/iec608705104/ManualTest.java
```diff
@@ -37,10 +37,10 @@ public static void main(String[] args) throws Exception {
                 throw new RuntimeException("Subscription not supported");
             }
 
-            plcConnection.subscriptionRequestBuilder().addChangeOfStateTagAddress("all", "*").addPreRegisteredConsumer("all", plcSubscriptionEvent -> {
+            plcConnection.subscriptionRequestBuilder().addChangeOfStateTagAddress("all", "*").setConsumer(plcSubscriptionEvent -> {
```
sruehl, line 40: shouldn't the consumer be set then as 3rd argument in addEventTagAddress following that logic? otherwise that would semantical not the same

sruehl (approved):

plc4j/drivers/firmata/src/test/java/org/apache/plc4x/java/firmata/readwrite/FirmataManualTest.java
chrisdutz: That was an example of how to register a handler for all tags of the request ... in that case it was only one, so yes ... there would be no difference to using the handler registation for an individual tag.

PR #1833, opened 2024-10-18
Fix coil read problem.

glcj:
Modbus driver does not read Coil & Discrete Inputs correctly.

1. It does not return Coil & Discrete Inputs
2. The bits are returned in order from top bit to bottom bit and must be inverted.

Coil writing needs to be tested.


chrisdutz (commented):
I implemented this in a way that it worked with the setup my coleagues were working with ... I do expect that these changes would break things with them.

For the different endianess we explicitly introduced the: 'default-payload-byte-order' configuration option.
Have you checked that none of the existing encoding options:
- BIG_ENDIAN
- LITTLE_ENDIAN
- BIG_ENDIAN_BYTE_SWAP
- LITTLE_ENDIAN_BYTE_SWAP
Help in your case?

chrisdutz (changes requested):
Having a more detailed look it seems you half-overrode the functionality I have in getReadBuffer in your adapted getResponseDataForTag method. Also removing the entire coil handing a bit higher in the class will 100% break things in the usecase that I tested things with.

glcj:
Hello,

Thanks for your time in checking,

I performed the tests with default configuration values ​​since the bit return order does not correspond to the order of the requested registers (not BYTES).

Eventually the modification corrects the reading of bit arrays in coils "0x00001:BOOL[48]". The driver  version 0.13.0-SNAPSHOT does not return the values ​​(at least until the tests performed last night).

Regarding the Inputs Register, it does return the bit array "1x00001:BOOL[48]" but not in the correct order.

If the people you point out are using the driver, they are probably only using the registers that do work. If you can discuss this with that team, that would be excellent.

Many applications use only registers.

I compared the traces against Applicom, CAS and ModbusPoll, so the protocol dialog is not the problem. I've already posted the screenshots on Slack with both types of records 0x0 and 1x0 requested and received in the correct order.

Please tell me which tests do not pass so I can compare apples with apples.

Thanks again,

chrisdutz:
So when working on the optimizer I added test-cases for the situations I was implementing ... it sort of seems as if you didn't run the testsuites. If I switch to your branch I see several of the tests failing. Please specially pay attention to the tests in the 
plc4x/plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/base/optimizer directory. I particularly modelled exactly the request structutes we were reading. So I guess the LittleEndianByteSwapTest is a perfect test for that usecase and it clearly states that your patch breaks this scenario.

Please pay a bit more attention to the tests that we have ... they are not for being skipped.

glcj:
Hi Chris,

I think my added value here is that I can test the driver against various devices and I can tell you again that Modbus for 0x and 1x functions has problems in the 0.13.0-SNAPSHOT development branch and a solution proposal is in this PR.

Here we are talking about bits and bytes, but in my mind I get an image of a valve that has CO2 on one side and NaOH on the other. Unit tests help, but my tests will be with real equipment. If I find any strange ones I will notify the team immediately, as I am actually doing.

Attached test with the Applicom driver as MODBUS server, the interesting thing about this driver is that it can vary the protocol depending on various TSX QUATUM models or standard Modbus.

Thank you for your support,





imagen


imagen

imagen

chrisdutz:
Well... The test is the data from a real world use case... I can have a more detailed look, but I am quite sure you're not setting the right endianess. I would expect your changes to not break my test case. Otherwise they break my usecase.

chrisdutz:
@glcj could you please simply whip up a unit test like the one I created for my test case? I guess only this way will we be able to implement something that works with both devices.

glcj:
Hi Chris,

I've already placed in the "fix/modbus" branch the test based on the frame returned by one of my devices and according to the pattern of the existing tests . (OpenModSim, Applicom and Odot gateway).

I tried to evaluate the frame of your reference test but none of the Modbus parsers could interpret it, I leave you in the notes the link of the tool (CAS doesn't interpret it either).

If you have the device model that you used to generate that pattern it would be interesting to have it as a general knowledge.

I'll be attentive to any observations,

Best regards,

chrisdutz:
So ... Even if we ignore the test that I built for the setup of my colleagues, I still have failing tests:
The ModbusTypAllDatatypesIT fails for the boolean register
Also the ManualModbusTCPDriverTest fails for the BOOL address

glcj:
Hi Chris,

To continue with the tests I ask you to work with a common device so we can compare apples with apples. I propose to you:

https://github.com/serhmarch

The results obtained with this Simulator are the same with the Applicom driver, ModbusPull, etc.

Or if you have another emulator of your choice that I can access, I will gladly download it.

We are going to run the driver without any type of BYTE exchange since the responses of these devices are standard. Regarding this point, I left you an observation in the test that I added.

It would also be helpful if you put the complete captures of the device under test.

Thank you for your time,

chrisdutz:
So ... having a more detailed look on your PR and the code you changed ... so in the ModbusOptimizer you removed the handling of the Coils and treat everything like a register ... however for a register one element consumes 2 bytes and a coil it consumes one bit. These "matchesRegister" and "matchesCoil" calls should produce quite different results. Not sure how it even works without it, admittedly.

I also think the main difference is: The optimizer is made for grouping multiple tags together and to reduce the number of requests sent to the PLC. In your case you simply read one tag, which is an array. I think I should probably have a look if there's anything special with that.

Right now merging this PR will effectively break the optimizer ... how about disabling/enabling it via configuration?

chrisdutz:
So I don't have the time to address this ... so if you ensure the test-suite is green, feel free to merge ... if things break for my colleagues (which I expect) I might get the time to fix it again, but with my current 4h/week ... no chance.

glcj:
Ok, I'll check everything again and verify the tests. With everything in green I'll perform the merge. If anything comes up I'll be on the lookout.

glcj:
The following changes were included in this modification.

1. Reading of a BOOL type from a HoldingRegister is enabled. In this particular case the HR must have a value of 0/1.
2. In the case of Coils and InputRegisters, they are always treated as BIG_ENDIAN, as specified by the standard.
3. In the case of optimized readings the value of the bit is interpreted directly from the acquired buffer and transferred as a short.
4. In all cases of bit readings in arrays or from an optimized call, the bits are returned in a consistent order.
5. All tests provided for the driver pass.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
```diff
@@ -69,22 +92,22 @@ public ManualModbusTCPDriverTest(String connectionString) {
     }
 
     public static void main(String[] args) throws Exception {
-        ManualModbusTCPDriverTest test = new ManualModbusTCPDriverTest("modbus-tcp://192.168.23.30");
+        ManualModbusTCPDriverTest test = new ManualModbusTCPDriverTest("modbus-tcp://10.10.1.200:10502?default-payload-byte-order=LITTLE_ENDIAN_BYTE_SWAP");
         test.addTestCase("holding-register:1:BOOL", new PlcBOOL(true)); // 0001
         test.addTestCase("holding-register:2:BYTE", new PlcBYTE(42)); // 2A
         //test.addTestCase("holding-register:3:WORD", new PlcWORD(42424)); // A5B8
-        test.addTestCase("holding-register:4:DWORD", new PlcDWORD(4242442424L)); // FCDE 88B8
+        test.addTestCase("holding-register:4:DWORD", new PlcDWORD(424244242L)); // 1949 7412
```
chrisdutz, line 99: Why change this to the smaller number? 4242442424 is a 32 bit unsigned value. Adding one more digit would make it a 33 bit number ... if you needed to go down to a 29-bit number, then I think there might be something wrong with the decoding.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
```diff
@@ -69,22 +92,22 @@ public ManualModbusTCPDriverTest(String connectionString) {
...
         test.addTestCase("holding-register:13:UINT", new PlcUINT(42424)); // A5B8
-        test.addTestCase("holding-register:14:DINT", new PlcDINT(-242442424)); // F18C 9F48
-        test.addTestCase("holding-register:16:UDINT", new PlcUDINT(4242442424L));// FCDE 88B8
+        test.addTestCase("holding-register:14:DINT", new PlcDINT(-242442424)); // 1949 7412
+        test.addTestCase("holding-register:16:UDINT", new PlcUDINT(424244242L));// FCDE 88B8
         test.addTestCase("holding-register:18:LINT", new PlcLINT(-4242442424242424242L));// C51F D117 B2FB B64E
         test.addTestCase("holding-register:22:ULINT", new PlcULINT(4242442424242424242L));// 3AE0 2EE8 4D04 49B2
         test.addTestCase("holding-register:26:REAL", new PlcREAL(3.141593F));// 4049 0FDC
-        test.addTestCase("holding-register:28:LREAL", new PlcLREAL(2.71828182846D)); // 4005 BF0A 8B14 5FCF
+        test.addTestCase("holding-register:28:LREAL", new PlcLREAL(2.71)); // 4005 BF0A 8B14 5FCF
```
chrisdutz, line 110: Why cut off the fractional digits? if you enter the hex value here: https://baseconvert.com/ieee-754-floating-point you get the number that I had in there (I think this was the precision that Java has). So if your driver returns only 2.71, I would assume that something's wrong.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/base/optimizer/LittleEndianByteSwapTest.java
```diff
@@ -183,9 +184,9 @@ public void testLittleEndianByteSwap() throws Exception {
         input.put("variable129",   new String[]{"coil:165:BOOL",             "false"});
         input.put("variable130",   new String[]{"coil:167:BOOL",             "false"});
         input.put("variable131",   new String[]{"coil:169:BOOL",             "false"});
-        input.put("variable132",   new String[]{"coil:171:BOOL",             "false"});
+        input.put("variable132",   new String[]{"coil:171:BOOL",             "true"});
```
chrisdutz, line 187: Why change this? This should not matter at all when reading and just requires me to change the program on my test-devices.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/base/optimizer/LittleEndianByteSwapTest.java
```diff
@@ -248,7 +249,8 @@ public void testLittleEndianByteSwap() throws Exception {
         Map<PlcReadRequest, BaseOptimizer.SubResponse<PlcReadResponse>> readResponses = new HashMap<>();
         readResponses.put(firstRequest, new BaseOptimizer.SubResponse<>(
             new DefaultPlcReadResponse(firstRequest, Map.of(
-                "coils0", new DefaultPlcResponseItem<>(PlcResponseCode.OK, new PlcRawByteArray(Hex.decodeHex("3060480c00c084000000000000000000000000000000")))))));
+                //"coils0", new DefaultPlcResponseItem<>(PlcResponseCode.OK, new PlcRawByteArray(Hex.decodeHex("3060480c00c084000000000000000000000000000000")))))));
+                "coils0", new DefaultPlcResponseItem<>(PlcResponseCode.OK, new PlcRawByteArray(Hex.decodeHex("10404004004004000000000000000000000000000044")))))));
```
chrisdutz, line 253: I assume the reason the data here changed, is because of the changed endianess in the connection string, right? In that case would I prefer us to have different tests for different encodings. Now it's impossible to switch between different types of PLCs ... in order to switch back to mine (if I updated the data in them) I would still need to change this code.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
glcj: This is because the application used (MonbusTools), his field editor does not allow me to enter the last digit. It does not affect the result since what is important is the order of the bytes.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
glcj: Here again it is because the ModbusTools tool and an OPC-DA client against the DirectLink driver apply scientific notation, so it is easier to read 2.71

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/base/optimizer/LittleEndianByteSwapTest.java
glcj: To validate that the bit position calculation within the optimization byte buffer is correct.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/base/optimizer/LittleEndianByteSwapTest.java
glcj: I used the bit pattern from the optimizer test, 175 coils, on two different devices and they returned the same byte sequence and according to the Modbus standard, so I used it as a reference.
I want to validate with other devices, but bit (coil) requests are always answered in BIG_ENDIAN, which is consistent with the standard, but I need to reconfirm this.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
chrisdutz: Yeah ... ok ... at least that's a reason I would accept ... possibly entering it as a signed integer would work? so probably some large negative number? (I just used some binary conversion tools to convert my initial value into a signed-integer value and the result is: -52524872) then we wouldn't be adjusting our tests to the bugs of a tool we use and would actually also test the boundaries of signed and unsigned integers.

plc4j/drivers/modbus/src/test/java/org/apache/plc4x/java/modbus/ManualModbusTCPDriverTest.java
chrisdutz: I intentionally defined the test data to test the range of the values ... so do I understand it correctly, that in the tool you can only define two decimal digits?

chrisdutz (approved):
Giving up on this ... merge if you like.

chrisdutz:
What's the status on this one?

chrisdutz:
Hmmm ... I guess I should probably close this .... the original base is completely out of date.

PR #1857, opened 2024-10-29
Reorganize page titles by 1 level

splatch:
Small re-organization of site titles so we get proper `<title>` tags in HTML output. Currently it fails to build.


chrisdutz (approved):
No real need to approve this ... you could locally build the site and have a look, or you merge the build and Jenkins will do that for you. If it breaks the site, we simply revert that commit. So I'd say: Go for it.

splatch:
It fails locally for me already, will try to adjust plugin versions, if it fails I'll lave it open.

PR #1878, opened 2024-11-08
fix(plc4j/drivers/s7): Delete 2 duplicate classes

jmeierhofer:
Delete 2 duplicate classes, the used ones are located in org.apache.plc4x.java.s7.readwrite.protocol.

image


sruehl:
@glcj why are those files in both locations? Fix looks fine, can't judge where they are supposed to be...

sruehl (approved):

glcj:
Hi,

S7Mux is actually an interface. Normally it should be in a ".api" package.

Its purpose is to be able to implement with a second instance of "S7MuxImpl", the case of four communication channels that Siemens achieves with the RECONNECT application [1].

A bit difficult to achieve, but oh well, that was the idea.

I don't see any problem with the modification.

1. https://mall.industry.siemens.com/mall/es/WW/Catalog/Product/6GK1716-0HB00-3AE0

sruehl:
K, ty for the PR

PR #1880, opened 2024-11-09
feat(site): Migrated the website to use Antora.

hutcheb:
Its in a position to be reviewed, still need to merge the changes from @splatch but it should be working.

To build it use.
mvn site

It doesn't use the build-tools repo.


chrisdutz:
This additional "src" module feels really a bit odd ... usually the "src" directoy is part of the plc4x-parent module and the convention is that sources for generating the site is located under "src/site". So how about creating a new "website" module and moving everything there? The current structure feels like you're mixing up things.

hutcheb:
@chrisdutz so it now has a website module and doesn't use the src folder.

PR #1902, opened 2024-11-19
feat(plc4j/ads): Add STRING and WSTRING to getDataTypeTableEntry.

fdupont-epsilia:
Add STRING and WSTRING to getDataTypeTableEntry.
 serializePlcValue, remove check directly in dataTypeTable, only use getDataTypeTableEntry.

### DIFF
--- a/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
+++ b/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
@@ -1198,7 +1198,6 @@ public class AdsProtocolLogic extends Plc4xProtocolBase<AmsTCPPacket> implements
     }
 
     protected byte[] serializePlcValue(PlcValue plcValue, String datatypeName) throws SerializationException {
-        // First check, if we have type information available.
         // Get the data type, allocate enough memory and serialize the value based on the
         // structure defined by the data type.
         Optional<AdsDataTypeTableEntry> dataTypeTableEntryOptional = getDataTypeTableEntry(datatypeName);
--- a/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
+++ b/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
@@ -1199,10 +1199,6 @@ public class AdsProtocolLogic extends Plc4xProtocolBase<AmsTCPPacket> implements
 
     protected byte[] serializePlcValue(PlcValue plcValue, String datatypeName) throws SerializationException {
         // First check, if we have type information available.
-        if (!dataTypeTable.containsKey(datatypeName)) {
-            throw new SerializationException("Could not find data type: " + datatypeName);
-        }
-
         // Get the data type, allocate enough memory and serialize the value based on the
         // structure defined by the data type.
         Optional<AdsDataTypeTableEntry> dataTypeTableEntryOptional = getDataTypeTableEntry(datatypeName);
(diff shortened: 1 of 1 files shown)


fdupont-epsilia:
Could this Pull Request be merged? Considering ADS works, and no modification in opcua.

```
test (21, ubuntu-latest)

[INFO] Protocols: ADS ..................................... SUCCESS [  2.192 s]
[INFO] PLC4J: Driver: ADS ................................. SUCCESS [ 12.046 s]

[INFO] PLC4J: Driver: OPC UA .............................. FAILURE [09:20 min]

Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test (default-test) on project plc4j-driver-opcua: 

2024-11-19 20:03:54,327 ERROR | nioEventLoopGroup-123-1 | o.a.plc4x.java.opcua.protocol.OpcuaSubscriptionHandle   | Invalid Tag Integer, subscription created without this tag 
2024-11-19 20:03:55,295  WARN | nioEventLoopGroup-123-1 | o.a.plc4x.java.opcua.protocol.OpcuaSubscriptionHandle   | Publish request of subscription 1 resulted in error reported by server 
java.util.concurrent.CompletionException: org.apache.plc4x.java.api.exceptions.PlcInvalidTagException: Integer invalid
```

chrisdutz:
Looks good to me ... haven't had the time to test it in my setup, but I wouldn't want to block merging this for that reason.

chrisdutz (approved):

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
```diff
@@ -1199,10 +1199,6 @@ protected CompletableFuture<PlcWriteResponse> multiWrite(PlcWriteRequest writeRe
 
     protected byte[] serializePlcValue(PlcValue plcValue, String datatypeName) throws SerializationException {
         // First check, if we have type information available.
```
sruehl: so maybe remove that comment too as it doesn't make sense anymore with the code below removed

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
```diff
@@ -1953,13 +1949,28 @@ protected Optional<AdsDataTypeTableEntry> getDataTypeTableEntry(String name) {
...
+                numBytes = adsDataType.getNumBytes();
+            }
+
             // !It seems that the dataType value differs from system to system,
             // !However, we never really seem to use that value, so I would say it doesn't matter.
             return Optional.of(new AdsDataTypeTableEntry(
-                128, 1, 0, 0, adsDataType.getNumBytes(), 0,
+                128, 1, 0, 0, numBytes, 0,
                 adsDataType.getValue(), 0, 0, 0,
-                adsDataType.name(), "", "",
```
sruehl, line 1962: is that semantical still the same? (`adsDataType.name()` vs. `name`)
ame)

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
chrisdutz: Good catch ... maybe a little bit of a comment to "first we try to read from the table" if nothing is found in TC2 systems the primitive types are omitted from the trable, we try if it's one of these primitive types, ... blah blah blah comments would leave some additional information for future coders.

fdupont-epsilia: Well, I keep it because it relates to « if(dataTypeTableEntryOptional.isEmpty()) { » below, but that's true it is no longer « 1st ».

plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/protocol/AdsProtocolLogic.java
chrisdutz: I think I remember that I had a look locally and this was also what I saw in the case of my TC3 devices Table.

chrisdutz: Our ENUM constants, that I defined in the early versions of the ADS driver match here ... I mean ... we did an enum value of based on the name ... so I would expect the other direction to produce the name name (Except for the sized strings ... )

fdupont-epsilia: No, "STRING(any lenght here)" is not the same as only "STRING"

chrisdutz: Possibly change it to that then? However I don't really think the information is ever read again ... but one would need to check that ... I think usually we use the "read by map-key" and don't really check the "getName" of the entry.

fdupont-epsilia: Sorry, there are delay in my reply. My reply was to sruehl question.
adsDataType.name() = "STRING"
name = "STRING(any length here)"
So they are not « semantical still the same » for STRING. For other base type (such as BOOL), they are the same. 
And we need "STRING(any length here)" for the rest of the code to work.

PR #1971, opened 2025-01-10
Adding comments

ayushjariyal:
Just adding comments on PLcEntityInterceptor.java file in Fallback section as i am just starting my open source journey

### DIFF
--- a/plc4j/tools/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityInterceptor.java
+++ b/plc4j/tools/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityInterceptor.java
@@ -520,6 +520,7 @@ public class PlcEntityInterceptor {
 
         // Fallback
         Object responseObject = response.getObject(sourceFieldName);
+        // Assuming getAll() returns a collection or related structure for validation
         if (clazz.isAssignableFrom(responseObject.getClass())) {
             return responseObject;
         }


plc4j/tools/opm/src/main/java/org/apache/plc4x/java/opm/PlcEntityInterceptor.java
```diff
@@ -520,6 +520,7 @@ static Object getTyped(Class<?> clazz, PlcReadResponse response, String sourceFi
 
         // Fallback
         Object responseObject = response.getObject(sourceFieldName);
+        // Assuming getAll() returns a collection or related structure for validation
```
sruehl, line 523: I don't understand the comment. Where does the `getAll()` come from?

ayushjariyal: Sorry, I am just starting my open-source journey. I have no intention to disturb you, but I want to ensure I am doing things correctly. Could you please help me in my journey?

ottlukas: First of all welcome the PLC4X Project :-). The question @sruehl  has is why do you call the function "getAll()" is the function not called "getObject()" ? So where is the function getAll() come from? Where is it called? Other than beeing a little bit more precise and answer the question we follow "community over code" so every contribution, comment, question, testing etc. is welcome.

PR #2016, opened 2025-02-19
fix(plc4j/ab-eth): AB-Ethernet Driver is not connecting in PLC Allen-Bradley SLC 5/05

andvasp:
The request is not being sent as requested in the protocol. According to the protocol it should send the data below.

    <name>Connection Request</name>
    <raw>01010000000000000000000000**0400050000**000000000000000000000</raw>

But it is sending like below:

    <raw>01010000000000000000000000**0000000000**00000000000000000000</raw>

The test is ok! It generates the correct request but when I execute it does not.

Running the test, it gets the information of [senderContext](https://github.com/apache/plc4x/blob/aaf78d9f45b4fa65dc048c87865560d046c4b47e/plc4j/drivers/ab-eth/src/main/generated/org/apache/plc4x/java/abeth/readwrite/CIPEncapsulationPacket.java#L207) from its own test xml file [ParserSerializerTestsuite](https://github.com/apache/plc4x/blob/d45cbf3990d43834bfe617271ef61703453bc42a/protocols/ab-eth/src/test/resources/protocols/abeth/ParserSerializerTestsuite.xml#L30-L29)

Should it be like this? I did not investigate others drivers if they behaves the same way

### DIFF
--- a/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/protocol/AbEthProtocolLogic.java
+++ b/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/protocol/AbEthProtocolLogic.java
@@ -54,6 +54,9 @@ public class AbEthProtocolLogic extends Plc4xProtocolBase<CIPEncapsulationPacket
     private static final Logger logger = LoggerFactory.getLogger(AbEthProtocolLogic.class);
     public static final Duration REQUEST_TIMEOUT = Duration.ofMillis(10000);
 
+    private static final List<Short> connectionRequestSenderContext = Arrays.asList((short) 0x00, (short) 0x04, (short) 0x00,
+        (short) 0x05, (short) 0x00, (short) 0x00, (short) 0x00, (short) 0x00);
+
     private static final List<Short> emptySenderContext = Arrays.asList((short) 0x00, (short) 0x00, (short) 0x00,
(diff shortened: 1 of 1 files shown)


chrisdutz:
Hi Anderson and welcome to PLC4X :-)

Unfortunately the AB-ETH protocol has not received any updates for almost 6 years. So happy if you can actually use it and are willing to put in a bit of work to improve it. 
Here most of the team simply don't have access to such devices, so we'd be delighted, to have you on board to improve things on that base.

Unfortunately I can't really say much about the protocol ... as I said, I built it 6 years ago and no longer have device access to keep it maintained.

If you say it doesn't work with AB devices without and it does with ... I would be willing to merge it.

andvasp:
Hi Christofer,

Thank you for your receptiveness and for reviewing it.

As you can see, the modifications are simple. I just change it to produce the request like described in [xml test case](https://github.com/apache/plc4x/blob/d45cbf3990d43834bfe617271ef61703453bc42a/protocols/ab-eth/src/test/resources/protocols/abeth/ParserSerializerTestsuite.xml#L30-L29) and double check comparing with other application requests.

After it is merged I will prepare another to reactivate the read mechanism. :)

Thanks.

andvasp:
Hi @chrisdutz!

Do I need to do anything else for the PR to be ready for merge?

Thanks.

chrisdutz:
Oh sorry.. I just started a new job today.... Was a bit hectic the list few days. I'll look into it ASAP.

chrisdutz:
Looked good to me.

PR #2028, opened 2025-03-03
feat(plc4j/opcua): Remove additional hostname resolution within OPC-UA

### LINKED ISSUE #1897: [Bug]: OPC-UA v0.13.0 struggles with servers without hostnames
### What happened?

We connect to an OPC UA server via a WAN connection where the hostname of the server is not present in our DNS, so we connect to it via IP address.  We had no problems connecting to this PLC using eg v0.12.0 however with v0.13.0 the connection is rejected, since the endpoints advertised on the server itself use the hostname (which is not visible in our network) instead of the IP address.  This is easily addressed by adding configuration parameter endpoint-host=<hostname> so now the URL (**URL 1**) looks like this:
```
opcua:tcp://10.3.1.13:4840?endpoint-host=ahostname
```
- however that leads to a new problem: v0.13.0 tries to connect to the server using ahostname instead of using the IP address 10.3.1.13.  This makes no sense: if we wanted to connect via hostname, we would use a much simpler URL (**URL 2**):
```
opcua:tcp://ahostname:4840
```
Thankfully, the code falls back to connecting via IP address, however it takes quite a bit of extra time, and a lot of errors and warnings are generated in the log.  IMHO this is illogical: there is already a very simple way to connect to OPC UA servers via hostname (via **URL 2**).  The only reason for the endpoint-host parameter that I am aware of is where the endpoint hostname and the connection address differ, with the point being to use the connection address (and not endpoint-host) to connect.

It should be possible to connect to an OPC UA server via IP address without any attempts to connect via hostname, ergo the format opcua:tcp://xxx?endpoint-host=hostname used by **URL 1** above should connect using xxx and not hostname.

### Version

v0.13.0

### Programming Languages

- [X] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [X] OPC-UA
- [ ] S7

splatch:
Closes #1897. Removed code which caused additional DNS lookups in favor of endpoint-port and endpoint-host options provided with plc4j 0.13.

### DIFF
--- a/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
+++ b/plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.plc4x.java.DefaultPlcDriverManager;
@@ -500,9 +501,11 @@ public class OpcuaPlcDriverTest {
     public void multipleThreads() throws Exception {
         class ReadWorker extends Thread {
             private final PlcConnection connection;
+            private final CountDownLatch latch;
 
-            public ReadWorker(PlcConnection opcuaConnection) {
+            public ReadWorker(PlcConnection opcuaConnection, CountDownLatch latch) {
                 this.connection = opcuaConnection;
+                this.latch = latch;
             }
 
             @Override
@@ -516,21 +519,24 @@ public class OpcuaPlcDriverTest {
                         PlcReadResponse read_response = read_request.execute().get();
                         assertThat(read_response.getResponseCode("Bool")).isEqualTo(PlcResponseCode.OK);
                     }
-
                 } catch (ExecutionException e) {
                     LOGGER.error("run aborted", e);
                 } catch (InterruptedException e) {
+                    LOGGER.error("thread interrupted", e);
                     Thread.currentThread().interrupt();
-                    throw new RuntimeException(e);
+                } finally {
+                    this.latch.countDown();
                 }
             }
         }
 
         class WriteWorker extends Thread {
             private final PlcConnection connection;
+            private final CountDownLatch latch;
 
-            public WriteWorker(PlcConnection opcuaConnection) {
+            public WriteWorker(PlcConnection opcuaConnection, CountDownLatch latch) {
                 this.connection = opcuaConnection;
+                this.latch = latch;
             }
 
             @Override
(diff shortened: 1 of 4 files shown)


plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
```diff
@@ -544,11 +550,13 @@ public void run() {
                         PlcWriteResponse write_response = write_request.execute().get();
                         assertThat(write_response.getResponseCode("Bool")).isEqualTo(PlcResponseCode.OK);
                     }
+                    this.latch.countDown();
                 } catch (ExecutionException e) {
                     LOGGER.error("run aborted", e);
+                    this.latch.countDown();
                 } catch (InterruptedException e) {
-                    Thread.currentThread().interrupt();
```
sruehl, line 550: this must stay in

plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
```diff
@@ -516,21 +519,24 @@ public void run() {
                         PlcReadResponse read_response = read_request.execute().get();
                         assertThat(read_response.getResponseCode("Bool")).isEqualTo(PlcResponseCode.OK);
                     }
-
+                    this.latch.countDown();
                 } catch (ExecutionException e) {
                     LOGGER.error("run aborted", e);
+                    this.latch.countDown();
                 } catch (InterruptedException e) {
-                    Thread.currentThread().interrupt();
```
sruehl, line 523: this must stay in

sruehl (changes requested):
not sure about  the implications of that change

sruehl:
other than that I don't have any objections

plc4j/drivers/opcua/src/test/java/org/apache/plc4x/java/opcua/OpcuaPlcDriverTest.java
splatch: Can you explain me why? We get this exception once thread is interrupted from outside, what is the point of calling interrupt again? Especially that in this place we are already out of the loop.

sruehl: because this ensures that the interrupt flag stays set https://github.com/joansmith/sonar-java/blob/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2142.html

sruehl: Out of curiosity: why did you remove the line?

splatch: Only one case for interrupted exception to happen is when test is terminated before completing 100 read or write attempts, long story short - its already dead. Because exception is caught outside of the loop thread is effectively terminated at the time exception is caught.

sruehl (approved):

PR #2040, opened 2025-03-15
feat(plc4j/ab-eth): Enable AB-Ethernet Driver reading

andvasp:
Hi everybody!

In this PR I am enabling the AB-Ethernet Driver reading. Also implemented the missing multiple tags reading.

About the transaction counter reinitialization, should not be set to 10 like the declaration?

@chrisdutz, Do you remeber why 10? Using one Opc Server it starts with 32. But 10 is working perfectly!

[transaction counter reinitialize](https://github.com/apache/plc4x/blob/a03d29f4063ace4f63e89b061d87f13a6ae7cc40/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/protocol/AbEthProtocolLogic.java#L117)
```
  private final AtomicInteger transactionCounterGenerator = new AtomicInteger(10);
...
  if (transactionCounterGenerator.get() == 0xFFFF) {
      transactionCounterGenerator.set(1);
  }
```

### DIFF
--- a/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/AbEthDriver.java
+++ b/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/AbEthDriver.java
@@ -85,6 +85,11 @@ public class AbEthDriver extends GeneratedDriverBase<CIPEncapsulationPacket> {
     }
 
     @Override
+    protected boolean canRead() {
+        return true;
+    }
+
+    @Override
     protected ProtocolStackConfigurer<CIPEncapsulationPacket> getStackConfigurer() {
         return SingleProtocolStackConfigurer.builder(CIPEncapsulationPacket.class, CIPEncapsulationPacket::staticParse)
             .withProtocol(AbEthProtocolLogic.class)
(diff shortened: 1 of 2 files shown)


chrisdutz:
Hehe... That's a relic from the old times in s7... There were several messages exchanged in order to establish the connection, so we simply had the counter start at 10...i guess there's absolutely no reason for this in ab-eth.

andvasp:
hehe...

What about the transaction counter reinitialization, the reinitialized value should not be the initial value on the variable declaration?

I tested using value 1 on the initialization and it works too. I checked the specification and it does not mention any constraint of the first value.

chrisdutz:
Feel free to change whatever you need.

chrisdutz:
Ups ... sorry ... the one button is directly next to the other one :-/

andvasp:
Hi @chrisdutz,

I changed the transaction counter initialization value.

Although it looks like a lot of modification, if you ignore whitespaces you are going to note that is not because I could remove a for statement.

Thanks,
Anderson.

andvasp:
Hi @chrisdutz,

I've seen that you are preparing for the next release. That is great!

Any chance to accept this PR to be included in the new release?

Thanks.
Anderson.

chrisdutz:
I however guess the release will probably still take a while (if I am to do it, as our new crowd-funding system is currently not really over-flowing with budget and I have stopped doing the housekeeping free of charge: https://opencollective.com/friends-of-toddy/projects/create-release-0130)

Issue #2229, opened 2025-08-27
[Bug]: Modbus - Number of threads repetitively increase at each reconnection.


frankpuppa:
### What happened?

While I was testing the stability of a modbus application based on plc4x, I faced the problem that on successful reconnection attempts the number of live threads keeps increasing continuously. Specifically, there is a pool as such: pool-N-thread-M (probably related to netty) which sits waiting and does not shut down. 
The same applies when using the CachedPlcConnectionManager
Connection string: modbus-tcp://127.0.0.1:9081?unit-identifier=1

Here is a sample code with springboot:
```java
 @Scheduled(fixedDelayString = "${scheduling.interval}") 
    public void update() {
        logger.info("Executing update method...")

        try {
            reconnect(false);

            if (plcConnection == null || !plcConnection.isConnected()) {
                logger.info("PLC connection is not alive, exiting...");
                return;
            }

            logger.info("OK");
        } catch (Exception e) {
            logger.error("Error while trying to read ... reconnecting", e);
        }
    }

      protected void reconnect(boolean error) {

        try {
            try {
                if (plcConnection != null && !plcConnection.isConnected()) {
                    logger.info("Reconnecting...");
                    // either with plcConnection.connect() or
                    // connectionManager.getConnection(plcConnectionString); There is a n increment
                    // of threads
                    // plcConnection = connectionManager.getConnection(plcConnectionString);
                    plcConnection.connect();
                    return;
                }
            } catch (Exception e) {
                logger.error("Error: {}", e.getMessage());
                return;
            }

            // Open a new one
            if (plcConnection == null) {
                logger.info("Connecting...");
                plcConnection = connectionManager.getConnection(plcConnectionString);
            }
        } catch (Exception ex) {
            logger.error("Error while trying to reconnect to the plc");
        }
    }
```
The function runs with a 10 seconds interval, in this time window I simulate the disconnection/reconnection of the cable.
I'm attaching a small video which shows this behavior more clearly. 

I understand that this is a corner case but it can lead to a waste of resources which I would like to avoid.

https://github.com/user-attachments/assets/003b4d80-e600-498a-b7bd-44c8697c4bd9

Thank you.

### Version

v0.13.0

### Programming Languages

- [ ] plc4c
- [ ] plc4go
- [x] plc4j
- [ ] plc4net
- [ ] plc4py

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] C-Bus
[...]

PR #2241, opened 2025-09-03
add changes to release resources when the channel becomes inactive

frankpuppa:
While I was looking into this issue I found out that there are two leaks. The first is related to the RequestTransactionManager and the other is from the NettyHashTimerTimeoutManager. As I was mentioning in the issue I opened last week, this leaks happens when the PLC goes offline and back online and the client tries to reconnecting to it in a cycle.

I was only able to test this on the modbus simulator which seems to work ok with the changes I made on this branch. 
Please provide me a feedback or any consideration. I am attaching also a small video which shows that the resources are freed as expected.

[Issue link](https://github.com/apache/plc4x/issues/2229)

https://github.com/user-attachments/assets/52d07125-cd87-47d7-9b96-e33f488a3039

### DIFF
--- a/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
+++ b/plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
@@ -136,14 +136,14 @@ public class Plc4xNettyWrapper<T> extends MessageToMessageCodec<T, Object> {
 
     @Override
     public void close(ChannelHandlerContext ctx, ChannelPromise promise) throws Exception {
-        logger.info("close.. context: {}", ctx.name());
+        logger.trace("close.. context: {}", ctx.name());
         super.close(ctx, promise);
         timeoutManager.stop();
     }
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
-        logger.info("channelInactive.. context: {}", ctx.name());
+        logger.trace("channelInactive.. context: {}", ctx.name());
         super.channelInactive(ctx);
         this.protocolBase.channelInactive(new DefaultConversationContext<>(this::registerHandler, ctx, authentication, passive));
         timeoutManager.stop();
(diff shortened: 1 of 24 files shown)


chrisdutz:
Hi Frank,
do I understand things correctly? The main difference is that you're overriding channelInactive in Plc4xNettyWrapper and forwarding that to the individual implementations?
Definitely looks sensible.

Chris

chrisdutz:
Could you please increase the log level for your additional log output to something "debug" or "trace"? "info" makes the drivers quite noisy.

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -136,10 +136,19 @@ public ExpectRequestContext<T> expectRequest(Class<T> clazz, Duration timeout) {
...
     @Override
     public void close(ChannelHandlerContext ctx, ChannelPromise promise) throws Exception {
+        logger.info("close.. context: {}", ctx.name());
         super.close(ctx, promise);
         timeoutManager.stop();
     }
 
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        logger.info("channelInactive.. context: {}", ctx.name());
```
chrisdutz: Please set this to "debug" or "trace"

plc4j/spi/src/test/java/org/apache/plc4x/java/spi/connection/DefaultNettyPlcConnectionTest.java
```diff
@@ -210,6 +213,13 @@ private void awaitIn(GateKeeper signal) {
                 throw new RuntimeException(e);
             }
         }
-    }
 
+        @Override
+        public void channelInactive(ConversationContext<Message> context) {
+            logger.info("On ChannelInactive");
```
chrisdutz, line 219: Please set this to "debug" or "trace"

chrisdutz (changes requested):

frankpuppa:
> Hi Frank, do I understand things correctly? The main difference is that you're overriding channelInactive in Plc4xNettyWrapper and forwarding that to the individual implementations? Definitely looks sensible.
> 
> Chris

Hello Chris,
Yes exactly, basically when the plc goes offline a channelInactive event is propagated. Within the nettywrapper I basically propagate the event in the implementations which can eventually shutdown the RequestTransactionManager and release the resources. I was able to test the functionality only for the ModbusTcpProtocol. I will fix the verbosity asap.

chrisdutz:
Hmmm ...  with the recent rewrite of all drivers and the SPI this PR now seems quite out of date ... is this still an issue?

PR #2273, opened 2025-09-22
Draft of SIP protocol mspec

splatch:


chrisdutz:
I see changes in the mspec grammar ... what are they for? The PR generally has no text .., what's the state of this PR?

splatch:
I had once an urge to look into telecom protocols and got SIP/SCTP and couple more to look at. For SIP which is mainly text protocol I faced issues with how the mspec parser treats strings. This PR changes string handling to allow multi line string sequences to pass.
Its a draft for a reason - it was a small experiment without intend to make it beyond PoC parser.

Issue #2335, opened 2025-11-10
[Feature Request]: Add polling interval for CHANGE_OF_STATE subscription type


kostinkirill:
### What would you like to happen?

PLC4j supports `CYCLIC`, `CHANGE_OF_STATE`, and `EVENT` subscription types. When we create a CYCLIC subscription we pass a parameter `pollingInterval` - interval, in which the tag should be polled. For CHANGE_OF_STATE we don't pass it as it's supposed that we will get the value as soon as it will be updated. But the problem is that the tag value might be updated too often, for example 100 times in a second, which creates a load on network and PLC. 

The idea of the improvement is to send `pollingInterval` when create `CHANGE_OF_STATE` to add an ability to limit the number of polling requests in case if PLC value is updated very frequently.

I checked with ADS PLC, it works. Not sure if other PLCs supports it

### Programming Languages

- [x] plc4j
- [ ] plc4go
- [ ] plc4c
- [ ] plc4net

### Protocols

- [ ] AB-Ethernet
- [x] ADS /AMS
- [ ] BACnet/IP
- [ ] CANopen
- [ ] DeltaV
- [ ] DF1
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] S7

chrisdutz:
I fully agree and I'm happy you brought it up as I just came across this yesterday.
Do you think adding variant with a Duration would be enough?

```java
        /**
         * Adds a new tag to the to be constructed request which should be updated as soon as
         * a value changes in the PLC.
         *
         * @param name       alias of the tag.
         * @param minInterval Minimum interval for updates (Limits the number of events for high frequency changes).
         * @param tagAddress tag address string for accessing the tag.
         * @param consumer   consumer for receiving update events for a given tag only.
         * @return builder.
         */
        PlcSubscriptionRequest.Builder addChangeOfStateTagAddress(String name, String tagAddress, Duration minInterval, Consumer<PlcSubscriptionEvent> consumer);

```
This was just one of the 4 methods involved, but I hope you get the idea.

kostinkirill:
> Do you think adding variant with a Duration would be enough?

@chrisdutz yes, I think so

Created a PR https://github.com/apache/plc4x/pull/2336

PR #2336, opened 2025-11-10
(feat): Add a minimum polling interval for CHANGE_ON_EVENT subscription type

kostinkirill:
Improvement for https://github.com/apache/plc4x/issues/2335

Added minimum polling interval for CHANGE_ON_EVENT subscription type

### DIFF
--- a/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcSubscriptionRequest.java
+++ b/plc4j/api/src/main/java/org/apache/plc4x/java/api/messages/PlcSubscriptionRequest.java
@@ -67,7 +67,7 @@ public interface PlcSubscriptionRequest extends PlcSubscriptionTagRequest {
          * Adds a new tag to the to be constructed request which should be polled cyclically.
          *
          * @param name            alias of the tag.
-         * @param tag           tag instance for accessing the tag.
+         * @param tag             tag instance for accessing the tag.
          * @param pollingInterval interval, in which the tag should be polled.
          * @return builder.
          */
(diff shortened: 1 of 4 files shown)


chrisdutz (changes requested):
Generally ok ... but in this case I guess we would need matching Duration-counterparts for:

- addChangeOfStateTagAddress (Without consumer)
- addChangeOfStateTag (Without consumer)
- addChangeOfStateTag (With consumer)

kostinkirill:
@chrisdutz added methods with dueration: 

- addChangeOfStateTagAddress (Without consumer)
- addChangeOfStateTagAddress (With consumer)
- addChangeOfStateTag (Without consumer)

chrisdutz:
I assume the next step would be to extend the ADS and possibly the OPC-UA driver to support this, right? Are you planning on continuing to work on this? I's prefer to mentor you through this as it's a bit like the "give a man a fish and he's fed for a day, teach him to fish and he's fed for life" sort of thing ;-)

chrisdutz:
The question now is: If a driver doesn't implement this Duration-based limiting ... should we throw an "unsupported" error, or should we silently forward it to the duration-less version and log a message?

kostinkirill:
> I assume the next step would be to extend the ADS and possibly the OPC-UA driver to support this, right? Are you planning on continuing to work on this? I's prefer to mentor you through this as it's a bit like the "give a man a fish and he's fed for a day, teach him to fish and he's fed for life" sort of thing ;-)

I believe ADS drive already supports it, I've already checked the behavior and `minInterval` is applied for CHANGE_ON_EVENT subscriptions and works as epxected. 
This the code where the `minInterval` is set up as `cycleTimeInMs` in `AdsAddDeviceNotificationRequest`
`AdsProtocolLogic#executeSubscribe:
```
return new AmsTCPPacket(new AdsAddDeviceNotificationRequest(configuration.getTargetAmsNetId(), configuration.getTargetAmsPort(),
                    configuration.getSourceAmsNetId(), configuration.getSourceAmsPort(),
                    0, getInvokeId(),
                    directAdsTag.getIndexGroup(),
                    directAdsTag.getIndexOffset(),
                    adsDataTypeTableEntry.getSize() * numberOfElements,
                    tag.getPlcSubscriptionType() == PlcSubscriptionType.CYCLIC ? AdsTransMode.CYCLIC : AdsTransMode.ON_CHANGE, // if it's not cyclic, it's on change or event
                    0, // there is no api for that yet
                    tag.getDuration().orElse(Duration.ZERO).toMillis()));
```
(the last argument)

As for OPCUA, I haven't checked it, but based on the code it should work as well

kostinkirill:
> The question now is: If a driver doesn't implement this Duration-based limiting ... should we throw an "unsupported" error, or should we silently forward it to the duration-less version and log a message?

I'd say drive implements it

chrisdutz:
Interesting ... wasn't expecting that to work ;-) 

But yeah ... having a look at the code, it actually makes sense. Cool ... so I guess this is ready to being merged ;-)

Thanks :-)

PR #2407, opened 2026-01-01
restore skip-code-generation property usage

hboutemy:
restore property usage introduced in 1ff58957f83cfa00bd033c27112f50ae04d41bec
but removed in #1172

necessary to rebuild the reference release published to Maven Central, that is build with the generation profile activated, which adds dependencies (even if provided)

see https://github.com/jvm-repo-rebuild/reproducible-central/pull/4800

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -971,6 +971,9 @@
           <groupId>org.apache.plc4x.plugins</groupId>
           <artifactId>plc4x-maven-plugin</artifactId>
           <version>${plc4x-code-generation.version}</version>
+          <configuration>
+            <skip>${skip-code-generation}</skip>
+          </configuration>
         </plugin>
 
         <plugin>


chrisdutz:
Admittedly I don't quite understand that change ... the plugin is only used (except for the mspec-tests) inside the update-generated-code profile. Also do we run the code generation in a preparation step before starting the release:prepare & repease:perform steps. As far as I can see it, only the "apache-release" profile is enabled when building the jars for maven central. (That might have been different in the past, but should be that way now)

Could you elaborate a bit on why this change is needed? Even if it's a trivial change, I can see "skip-code-generation" is being defined and set to false, but it's not used anywhere, so removing the property all together seems to make more sense.

hboutemy:
perhaps I did a mistake when deducting root cause of the difference https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/plc4x/plc4x/plc4x-parent-0.13.1.diffoscope

the fact is that for now, I can't find a way to rebuild plc4x 0.13.1 and get the same output (beyond the small `develocity-workspace-id` difference)

you know the build better than me, perhaps the root cause is something else that will be easier for you to analyze

chrisdutz:
It looks a bit like one was built with "with-dotnet" and "with-python" enabled and one without. But no idea why the mvn:commons-io/commons-io/2.20.0 is added.

hongjinlin:
I agree that restoring the `skip-code-generation` property is useful for local builds. 

However, like Chris mentioned, the appearance of `commons-io` is unexpected.

chrisdutz:
Sorry, I won't be able to further look into this due the next 11 days, as I'm on vacation and not taking any of my euro equipment with me.

chrisdutz:
Hmmm ... thinking more about it ... usually we need to explicitly enable code-generation ... never disable it?

PR #2458, opened 2026-02-20
Fix/opc ua fine tuning

chrisdutz:
Please review :-)

These were changes needed to get the manual test to work with the OPC-UA server of my S7-1500 device.


sruehl:
seems like you did not adjust the Golang version? Why was it required to go to a manual field? Sometimes it is a bit hard to find the static helper in the mega diff :D

chrisdutz:
In order to read some of the time fields I needed to add this alternative ... but yeah ... didn't know we had an OPC-UA driver in Go ... sorry for that (I stopped building Go as it takes for ever every time)

sruehl:
no worries, the go build where broken due to the invoker messup for a while so maybe that didn't jump directly to the eye. Since I repaired it red should really mean build fail 🗡️

plc4j/spi/src/main/java/org/apache/plc4x/java/spi/Plc4xNettyWrapper.java
```diff
@@ -198,7 +198,9 @@ protected void decode(ChannelHandlerContext channelHandlerContext, T payload, Li
                     logger.trace("Failure while processing payload {} with handler {}", message, registration, e);
                     BiConsumer biConsumer = registration.getErrorConsumer();
                     if(biConsumer != null) {
-                        biConsumer.accept(message, e);
+                        // Use the original payload rather than the transformed message,
+                        // as the error consumer expects the original message type.
+                        biConsumer.accept(payload, e);
```
splatch, line 203: Error handlers are "global", hence by introducing this change you may break some drivers in situations which we do not test.

protocols/opcua/src/main/xslt/opc-manual.xsl
```diff
@@ -225,7 +225,7 @@
             [typeSwitch encodingMask.xmlBody, encodingMask.binaryBody
                 ['false', 'true' BinaryExtensionObjectWithMask
                     [implicit int 32 bodyLength 'body == null ? 0 : body.lengthInBytes']
-                    [simple ExtensionObjectDefinition('extensionId') body]
+                    [manual ExtensionObjectDefinition body 'STATIC_CALL("parseExtensionObjectBody", readBuffer, extensionId, bodyLength)' 'STATIC_CALL("serializeExtensionObjectBody", writeBuffer, body)' 'body == null ? 0 : body.lengthInBits']
```
splatch, line 228: I understand reason for the change, but better move  needed logic and `UnknownExtensionObject` to mspec, so it will be portable across languages.

plc4j/drivers/opcua/src/main/java/org/apache/plc4x/java/opcua/readwrite/UnknownExtensionObject.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.plc4x.java.spi.generation.SerializationException;
+import org.apache.plc4x.java.spi.generation.WriteBuffer;
+import org.apache.plc4x.java.spi.generation.WriteBufferBoxBased;
+
+/**
+ * Represents an ExtensionObject whose type is not known to the parser.
+ * Stores the raw body bytes so they can be interpreted later based on context
+ * (e.g., the tag's dataType hint).
+ */
+public class UnknownExtensionObject extends ExtensionObjectDefinition {
```
splatch, line 32: Please move this definition to mspec.

splatch (changes requested):
Some of changes are unrelated to opc-ua. Some other changes affect existing logic. Please revisit these.

PR #2459, opened 2026-02-22
chore: Extended the PlcConnection metadata to include the driver and …

chrisdutz:
…the transport code and names.


chrisdutz:
Guess I'm gonna play with this in my own drivers first and bring one PR back as soon as I've got everything in the shape that fits all my needs.

PR #2382, opened 2025-12-08
SPI 3 Refactoring

chrisdutz:
In this refactoring, we are replacing the existing Netty-based core into a set of lightweight sub-modules without requiring Netty at all and with as little third party dependencies as possible.

This will allow using our code-generation framework independently from building PLC4J drivers and allow porting PLC4X to other languages as we no longer rely on external libraries for which counterparts need to be integrated for new languages.

PR #2471, opened 2026-02-27
fix(plc4j/eip): NullPointerException on decodeSingleReadResponse.

andvasp:
Hi everybody!

I am facing NPE at [EipProtocolLogic:::decodeSingleReadResponse](https://github.com/apache/plc4x/blob/cabd026ac1f00c39d09865a0e66fddf25da639f2/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/protocol/EipProtocolLogic.java#L814) because the PlcResponseCode == "INTERNAL_ERROR
```
╔═CipService════════════════════════════════════════════╗
║╔═response╗╔═service╗╔═CipReadResponse════════════════╗║
║║ b1 true ║║0x4c 76 ║║╔═reserved╗╔═status╗╔═extStatus╗║║
║╚═════════╝╚════════╝║║ 0x00 0  ║║0x05 5 ║║  0x00 0  ║║║
║                     ║╚═════════╝╚═══════╝╚══════════╝║║
║                     ╚════════════════════════════════╝║
╚═══════════════════════════════════════════════════════╝
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.plc4x.java.eip.readwrite.CIPData.getDataType()" because the return value of "org.apache.plc4x.java.eip.readwrite.CipReadResponse.getData()" is null
	at org.apache.plc4x.java.eip.base.protocol.EipProtocolLogic.decodeSingleReadResponse(EipProtocolLogic.java:814)
	at org.apache.plc4x.java.eip.base.protocol.EipProtocolLogic.lambda$readWithoutMessageRouter$5(EipProtocolLogic.java:509)
	at java.base/java.util.function.Consumer.lambda$andThen$0(Consumer.java:65)
	at org.apache.plc4x.java.spi.Plc4xNettyWrapper.decode(Plc4xNettyWrapper.java:194)
	at io.netty.handler.codec.MessageToMessageCodec$1.decode(MessageToMessageCodec.java:67)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:91)
	at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:120)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442)

This pull request just move the type and data to inside the "if" where PlcResponseCode is OK. The data moved is just necessary inside the if to create the plcValue. So it is ok to move the code.

Instead of throwing an exception, the system will report the response code INTERNAL_ERROR and plcValue will be null.

PlcValue null value already can happen on parsePlcValue. So not prejuize to the current logic.

Best regards,
Anderson.

### DIFF
--- a/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/protocol/EipProtocolLogic.java
+++ b/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/protocol/EipProtocolLogic.java
@@ -811,9 +811,9 @@ public class EipProtocolLogic extends Plc4xProtocolBase<EipPacket> implements Ha
         CipReadResponse resp = (CipReadResponse) p;
         PlcResponseCode code = decodeResponseCode(resp.getStatus());
         PlcValue plcValue = null;
-        CIPDataTypeCode type = resp.getData().getDataType();
-        ByteBuf data = Unpooled.wrappedBuffer(resp.getData().getData());
         if (code == PlcResponseCode.OK) {
+            CIPDataTypeCode type = resp.getData().getDataType();
+            ByteBuf data = Unpooled.wrappedBuffer(resp.getData().getData());
             plcValue = parsePlcValue((EipTag) tag, data, type);
         }
         PlcResponseItem<PlcValue> result = new DefaultPlcResponseItem<>(code, plcValue);


andvasp:
Hi @chrisdutz!

May you take a look in this simple PR?

Thanks,
Anderson.

andvasp:
Hi @chrisdutz !

I have done an initial test using your work in [refactor/spi3](https://github.com/apache/plc4x/tree/refactor/spi3) and looks like this is not happen anymore. I will do more tests and report it!

In your e-mail [[SPI3] New CAN, AB-ETH and C-Bus driver](https://lists.apache.org/thread/lmmn8r85ls9t3hl73r1n8vmzz2xp31xd). You have done the comment below: 
"About AB-ETH and C-Bus I sort of feel we should probably drop them from the repo until we have people using them"

I am using the AB-ETH, inclusive I have done a contributtion to it on #2040. I will also do tests on it too!

Sorry to report the test here. What is the best way to report the test on the refactor/spi3?

The configuration at [s7Driver](https://github.com/apache/plc4x/blob/refactor/spi3/plc4j/drivers/s7/src/main/resources/META-INF/services/org.apache.plc4x.java.api.PlcDriver) file is incorrect. Perhaps some tests you are doing.

Thanks,
Anderson.

chrisdutz:
Oh wow ... thank you so much for providing feedback at all ;-) I'm happy to hear people are testing things and also that they are using our drivers. 

I was sort of expecting to have the bug reports come in as soon as I merge it back to develop because I thought: Nobody is really going to test things till it's in develop. You prove the opposite.

You could leave comments on the PR here ... this way the NLNet folks see that people are using and testing the new driver.

andvasp:
It is no longer necessary since #2382 fixed the problem.

PR #2535, opened 2026-04-24
fix(plc4py/umas): resolve Python 3.12+ SyntaxWarnings in UmasTag

Abdylreshit:
## Summary

Fixes three `SyntaxWarning` messages emitted by Python 3.12+ in `plc4py/drivers/umas/UmasTag.py`:

1. Invalid escape sequence `\[` in the regex `_ADDRESS_PATTERN` — converted the string literal to a raw string.
2. Two occurrences of `len(...) is not 0` — replaced with `!= 0`. Comparing `int` with `is`/`is not` is unreliable (works only by CPython small-int caching) and the behavior is deprecated.

## Test plan
- [x] `pytest` passes (77 passed, 36 xfailed — same as before, now with 0 warnings)
- [x] No behavioral changes — both fixes are semantics-preserving

### DIFF
--- a/plc4py/plc4py/drivers/umas/UmasTag.py
+++ b/plc4py/plc4py/drivers/umas/UmasTag.py
@@ -27,7 +27,7 @@ from plc4py.spi.messages.PlcRequest import TagBuilder
 
 class UmasTag(PlcTag):
     _ADDRESS_PATTERN: str = (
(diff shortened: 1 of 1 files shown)


chrisdutz:
Any of the python folks able and willing to have a look? @ottlukas @hutcheb ?

ottlukas (commented):
This is a minor regex and syntax change. FIne for me. Thanks for catching :-)

PR #2542, opened 2026-04-29
fix(plc4j/drivers/s7): stop S7HMuxImpl from leaking direct buffers (fixes #2248)

michele-tramonti:
## Summary

Fixes #2248 — direct memory leak in the S7 driver that exhausts `MaxDirectMemorySize` on long-running applications (`OutOfMemoryError: Cannot reserve N bytes of direct buffer memory`).

`S7HMuxImpl` is installed on three pipelines (`EmbeddedChannel` + primary/secondary TCP), and its `encode()` was both forwarding the outbound message to the active TCP channel and pushing a freshly-copied direct `ByteBuf` into the `MessageToMessageCodec` out list. That second copy ends up in the `EmbeddedChannel`'s outbound queue, which nothing in plc4j ever drains. Result: every sent S7 message leaked one direct buffer, leading to the OOM described in the issue (1-2 weeks to reproduce in production).

The fix replaces the second copy with the `Unpooled.EMPTY_BUFFER` singleton — Netty still requires at least one element in the out list (otherwise `MessageToMessageEncoder` throws `EncoderException`), but the singleton allocates zero direct memory and its `release()` is a no-op. The forward to the active TCP channel (`outBB.copy()` + `writeAndFlush`) is unchanged, so wire behaviour is identical.

## Why `copy()` and not `retainedDuplicate()` for the TCP forward

`copy()` is kept on purpose: future contributors who modify `S7HMuxImpl` (e.g. add another consumer, change the failover logic, insert handlers in the TCP pipeline) shouldn't have to reason about shared refcounts and concurrent reads of the same backing memory. The cost of one `copy()` per S7 request is negligible compared to the safety margin.

## Regression test

`S7HMuxLeakTest` wires `S7HMuxImpl` exactly like `S7HPlcConnection` does (one `EmbeddedChannel` as the logical pipeline, one as the primary TCP channel), runs 2000 outbound writes and asserts:

1. each message is forwarded exactly once to the TCP side,
2. the `EmbeddedChannel`'s outbound queue does not accumulate non-empty `ByteBuf`s.

Verified locally: the test fails on the unfixed code with `expected: <0> but was: <2000>` (~512 KB of leaked buffers, exactly matching the synthetic payload), and passes with the fix.

## Test plan

- [x] `mvn -pl :plc4j-driver-s7 -P with-java test -Dtest=S7HMuxLeakTest` passes with the fix
- [x] Same test fails deterministically on the unfixed code path (confirms the regression test is meaningful)
[...]

Issue #2248, opened 2025-09-06
[Bug]: S7 protocol leaks memory


michele-tramonti:
### What happened?

i'm using plc4j version 0.13.0, connecting to S7_400.
sometimes (I'm not able to identify the specific reason or pattern) i see this error in log:
```[2025-09-04 15:51:56] [info] 2025-09-04 15:51:56.923 ERROR 1894504 --- [tLoopGroup-24-1] io.netty.util.ResourceLeakDetector       : LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
[2025-09-04 15:51:56] [info] Recent access records:
[2025-09-04 15:51:56] [info] Created at:
[2025-09-04 15:51:56] [info] #011io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:410)
[2025-09-04 15:51:56] [info] #011io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
[2025-09-04 15:51:56] [info] #011io.netty.buffer.UnsafeByteBufUtil.copy(UnsafeByteBufUtil.java:454)
[2025-09-04 15:51:56] [info] #011io.netty.buffer.PooledUnsafeDirectByteBuf.copy(PooledUnsafeDirectByteBuf.java:216)
[2025-09-04 15:51:56] [info] #011io.netty.buffer.AbstractByteBuf.copy(AbstractByteBuf.java:1194)
[2025-09-04 15:51:56] [info] #011org.apache.plc4x.java.s7.readwrite.protocol.S7HMuxImpl.encode(S7HMuxImpl.java:136)
[2025-09-04 15:51:56] [info] #011org.apache.plc4x.java.s7.readwrite.protocol.S7HMuxImpl.encode(S7HMuxImpl.java:1)
[2025-09-04 15:51:56] [info] #011io.netty.handler.codec.MessageToMessageCodec$2.encode(MessageToMessageCodec.java:85)
[2025-09-04 15:51:56] [info] #011io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:90)
[2025-09-04 15:51:56] [info] #011io.netty.handler.codec.MessageToMessageCodec.write(MessageToMessageCodec.java:130)
[...]

PR #2543, opened 2026-04-29
fix(plc4j/drivers/s7): release direct ByteBufs allocated for alarm-query reassembly

michele-tramonti:
## Summary

Second direct-memory leak I noticed while auditing #2248 (and its fix #2542). Same family of bug, different code path.

`S7ProtocolLogic.decodeEventSubscriptionResponse()` allocates two direct `ByteBuf`s when handling `S7PayloadUserDataItemCpuFunctionAlarmQueryResponse`:

\`\`\`java
ByteBuf buffer = Unpooled.directBuffer(items.getItems().length * 2);
ByteBuf rxBuffer = Unpooled.directBuffer(items.getItems().length * 2);
\`\`\`

They are populated, then drained via \`ByteBufUtil.getBytes(rxBuffer)\` (which copies the bytes into a Java array) and the method returns. Neither buffer is ever \`release()\`d. \`ByteBufUtil.getBytes\` does not own the buffer and does not release it.

On a PLC that emits alarms regularly, every alarm event leaks two direct buffers. On long-running applications this slowly exhausts \`MaxDirectMemorySize\` and triggers \`OutOfMemoryError: Cannot reserve N bytes of direct buffer memory\` — the same symptom as #2248, different root cause.

## Fix

Wrap the existing logic in a `try/finally` so both buffers are released on every path, including when the inner parsing block (or any of the `loopFuture.get()` calls) propagates an exception.

No change in behaviour or wire format — the buffers are local to the method and only used for reassembly before the bytes are copied out via `ByteBufUtil.getBytes`.

## Test plan

- [x] `mvn -pl :plc4j-driver-s7 -P with-java test` passes

### DIFF
--- a/plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
+++ b/plc4j/drivers/s7/src/main/java/org/apache/plc4x/java/s7/readwrite/protocol/S7ProtocolLogic.java
@@ -771,70 +771,75 @@ public class S7ProtocolLogic extends Plc4xProtocolBase<TPKTPacket> {
 
             ByteBuf buffer = Unpooled.directBuffer(items.getItems().length * 2);
             ByteBuf rxBuffer = Unpooled.directBuffer(items.getItems().length * 2);
-            buffer.writeBytes(items.getItems());
-
-            if (itemparameter.getLastDataUnit() == 1) {
-                short loop = 0xff;
-                CompletableFuture<S7MessageUserData> loopFuture;
-                S7MessageUserData msg;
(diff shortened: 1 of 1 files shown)


chrisdutz:
Hi Michele,

thanks for the PR ... however I should mention that in the SPI3 PR I'm currently completely rewriting all of our drivers to completely work without Netty and everything that was making our life miserable. 

Chris

chrisdutz:
Could you please double-check now? I've just merged the SPI3 refactoring.

chrisdutz:
If I won't receive feedback here, I'll simply consider this done.

Issue #2585, opened 2026-06-04
[Feature Request]: Contribution proposal: read-only SLMP (MELSEC-Communication 3E) driver


LivingLikeKrillin:
Following up on #1728 (the answer there was essentially "no plans, but contributions welcome") — I'd like to build this.

I've previously implemented the MC protocol family in production — the 3E frame in both ASCII and binary codings — so the protocol itself is well-trodden ground for me; the genuinely new part for me is PLC4X's `mspec`/SPI conventions. I'd implement strictly from Mitsubishi's publicly available *MELSEC Communication Protocol Reference Manual* (SH-080008), which documents the 3E/4E frames, the device-code table, and the batch/random-read commands — so there's no dependency on member-restricted CLPA material.

To keep it reviewable, I'd start with a deliberately small **read-only** PoC and grow it in stages:

- **v0 (first PR):** 3E binary frame, read-only, Batch Read (`0x0401`, word units), `D` device, types INT/WORD/DINT/REAL, single request — plus a ParserSerializer test-suite so the wire layer is verifiable **without hardware**.
- **then, in stages:** more devices (W/R/M/X/Y/B) → bit units → Random Read (`0x0403`) with an optimizer (batch-vs-random selection, max-points splitting) → ASCII coding → 4E frame (serial-number correlation) / write / iQ-R extended addressing.

On naming and structure: I'd call the driver `slmp` (the open CLPA spec name, which covers the MC protocol — happy to switch to `mc`/`melsec` if you'd prefer), laid out mirroring the **modbus** driver: stateless TCP request/response, a `base` module plus thin per-coding modules, and a `discriminatedType` ADU switching frame-type/coding, which maps cleanly onto 3E/4E × binary/ASCII. The device-addressing **tag layer** I'd model on **s7** (`MemoryArea` + bit/word transfer-size + offset), since SLMP's device-code matrix (D/M/X/Y/W/B/R, per-device radix, bit/word) is structurally much closer to s7 than to modbus's flat register model. The later read optimizer would draw on s7's `S7Optimizer`/`S7BlockReadOptimizer`.

Before I open the first PR, two things I'd rather confirm up front:

1. **Placement** — does `protocols/slmp` (mspec) + `plc4j/drivers/slmp` work, or would you rather it incubate elsewhere first?
2. **Granularity** — is a staged, read-only-first PoC PR an acceptable way to start, or would you rather see a larger initial scope?

[...]

PR #2597, opened 2026-06-13
Add SLMP (MELSEC Communication 3E) protocol module (read-only wire layer)

LivingLikeKrillin:
﻿First increment of the SLMP driver proposed in #2585 — the codegen-layer protocol module only (`protocols/slmp`), opened as a draft per the Slack suggestion so the work is visible early.

### What''s in it
- 3E binary frame (request/response discriminated on the subheader byte), little-endian, read-only
- Commands: Batch Read (`0x0401`), Random Read (`0x0403`), Batch Read Multiple Blocks (`0x0406`), word units
- Root type declares the new-SPI encoding defaults
- ParserSerializer vectors derived from the worked examples in the public Mitsubishi reference manual SH-080008 (sections 8.1 / 8.3 / 8.4)

### Validation so far
- `mvn -pl protocols/slmp test` green on current develop (post-SPI3 merge): the mspec parses and all type references resolve
- `0x0401` / `0x0403` cross-checked against pymcprotocol (independent MIT-licensed client): its request bytes are byte-identical to the vectors except the monitoring-timer field, and it decodes the responses to the manual''s exact values
- `0x0406`: pymcprotocol has no multi-block API, so the request vector rests on the manual''s worked example plus an independent re-encode check — flagging this honestly
- The ParserSerializer suite is not yet executed by any build module — like the other protocols, it runs from the (upcoming) driver module; until then the in-module test validates that the mspec parses and all type references resolve
- Byte round-trip through generated code will come with the driver module

### Next (not in this PR)
- `plc4j/drivers/slmp` on the new SPI (slmp-tcp; single in-flight request since the 3E frame has no serial number)
- Bit units, more devices, write commands, ASCII mode, 4E — per the staged roadmap in #2585

---
Developed with AI assistance; every wire-layer byte was verified against SH-080008, with the batch and random reads additionally cross-checked against pymcprotocol as noted above.

### DIFF
--- a/protocols/pom.xml
+++ b/protocols/pom.xml
@@ -59,6 +59,7 @@
     <module>profinet</module>
     <module>s7</module>
     <module>simulated</module>
+    <module>slmp</module>
     <module>socketcan</module>
     <module>umas</module>
   </modules>
--- a/protocols/slmp/pom.xml
+++ b/protocols/slmp/pom.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.plc4x</groupId>
+    <artifactId>plc4x-protocols</artifactId>
+    <version>0.14.0-SNAPSHOT</version>
+  </parent>
+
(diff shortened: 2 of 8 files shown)


chrisdutz:
I've checked out your PR and created a dummy slmp driver module where the code is generated from your mspec and added a ParserSerializerTest to use the generated code with your xml ... I had to set it to fix one general issue: in your initial XML you had "deviceCode" modelled as:

`    <deviceCode>D</deviceCode>
`
However it's an enum with a numerical value, so correct would be: 

```
                <deviceCode>
                  <SlmpDeviceCode dataType="uint" bitLength="8" stringRepresentation="D">168</SlmpDeviceCode>
                </deviceCode>

```
This was the only issue I found in getting the driver's types, parsers and serializers generated and the test-suite running.

You want me to send you my changes or do you want to learn it yourself?

LivingLikeKrillin:
Thanks for taking the time to check it out and for wiring up the generated-driver test run — much appreciated.

Good catch on the `deviceCode` representation. I actually ran into the exact same thing while building the driver on my (separate, stacked) branch, and had already converted the test vectors to the canonical enum form (`<SlmpDeviceCode … stringRepresentation="D">168</SlmpDeviceCode>`). I've just pushed that fix to this PR, so no need to send yours — but thank you for the offer.

On the bigger picture: this PR is intentionally the protocol-only module (mspec + generated wire layer), which is why the test suite isn't exercised here yet. The Java driver that actually runs these vectors is the next, separate PR, stacked on top of this one — I'll open it once this merges so its diff stays clean. Happy to adjust if you'd prefer different sequencing.

hutcheb (approved):
Looks good.

chrisdutz:
Ok ... well ... as the PR says: Protocol Module I think it's ok to merge ... could you however do me a favor and sign an ICLA with Apache? This is something all contributors with more than minor contributions have to sign ... especially if you want to become a long time contributor or even committer here.
 
https://www.apache.org/licenses/icla.pdf

Then we've got everything in the right form and I'm happy to merge this PR

LivingLikeKrillin:
Thanks, Chris! I've signed and submitted my ICLA to secretary@apache.org. Happy to sign it — I'd like to keep contributing to PLC4X for the long haul.

chrisdutz:
Awaiting the confirmation and will hit merge as soon as that's confirmed.

chrisdutz:
We just got a confirmation about a newly signed ICLA ... but that name on it leaves no way for me to match with your user account here ... could you please send me a private message to cdutz@apache.org and tell me the name on your ICLA so I can formally match the two?

PR #2606, opened 2026-06-18
plc4j(ab-eth): Fix read tag after refactoring/spi3

andvasp:
Hi @chrisdutz!

When trying to read tag values, the system is getting a timeout error. The problem occurs because the calculateTotalMessageSize is getting a big value. As a result, the driver keeps waiting for incoming data that will never arrive, since the actual data transfer has already finished. 

Note: I considered whether this behavior should be configured. However, since the rest of the reading mechanism works fine without extra configuration, a direct fix in the size calculation seems to be the correct approach.

Before refactoring/spi3 it is working properly. I tried to find similar logic in the old version but It looks like it does not exist. So I believe the old mecanism approach considers that the received data will arrive in just one shot, right?

Thanks,
Anderson.

### DIFF
--- a/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/AbEthMessageCodec.java
+++ b/plc4j/drivers/ab-eth/src/main/java/org/apache/plc4x/java/abeth/AbEthMessageCodec.java
@@ -68,8 +68,8 @@ public class AbEthMessageCodec extends MessageCodecBase<CIPEncapsulationPacket> 
 
     @Override
     protected int calculateTotalMessageSize(byte[] header, int availableBytes) throws MessageCodecException {
-        // Little-endian unsigned 16-bit at offset 2
-        int payloadLength = (header[LENGTH_OFFSET] & 0xFF) | ((header[LENGTH_OFFSET + 1] & 0xFF) << 8);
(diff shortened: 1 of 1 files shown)


chrisdutz:
Looking at your change it looks as if this is a Little-Endian vs. Big-Endian thing ... However looking a bit deeper into this, it seems the EIP is generally LE ... I wonder why things worked in my case. I'll look into this.

chrisdutz:
Just noticed that it's not the EIP driver, but the AB-ETH ... sorry for the mis-understanding ... so you state that all is working with this fix? Then I'll simply merge it as I have no other way to validate it and the code changes look sound.

andvasp:
Yes. It is working fine. If you check in the [ab-eth.mspec](https://github.com/apache/plc4x/blob/develop/protocols/ab-eth/src/main/resources/protocols/abeth/ab-eth.mspec) you will notice that it specify big-endian.

I was responsible for enabling reading of this driver through #2040, Probably I am the only one using it! :)

Thanks!

chrisdutz:
Yeah ... might be ... would really be cool, if we knwo who's using what ...

Aaah ... I had it little endian and you made it big endian to match ... perfect .. thanks for spotting this and fixing it ;-)

PR #2630, opened 2026-07-07
fix(plc4j/ads): connect handshake honors configured AMS ports instead of hardcoded RUNTIME_SYSTEM_01

### LINKED ISSUE #2629: [Bug]: ADS connect handshake ignores the configured target/source AMS port
### Description

The ADS driver's connection handshake sends its probe requests to a **hardcoded** AMS port
`DefaultAmsPorts.RUNTIME_SYSTEM_01` (851), with a hardcoded source AMS port `800`, instead of the
`target-ams-port` / `source-ams-port` from the connection string. Only the handshake is affected -
the actual read/write path correctly uses `getConfiguration().getTargetAmsPort()`.

As a result, **any TwinCAT system who has no PLC runtime on AMS port 851 cannot be connected to**,
even when `target-ams-port` is set correctly. This includes a second PLC project (852), a project
that was assigned a non-default port, and any multi-runtime setup where several PLC apps run in
parallel (852, 853, ...) where we do not want to handshake on 851. The connection fails during the
handshake - before any user read/write - with ADS error `0x6` ("target port not found"), because 
the `ReadDeviceInfo` is delivered to port 851 where no runtime is listening. 

I am unaware of the intention hardcoding 851 and assume it is unintended to not cover the uscase 
of supporting parallel PLC runitmes that do not include 851 among their used ports.

### Affected code

`plc4j/drivers/ads/.../AdsTcpConnection.java`:

- `doConnectHandshake()` - three requests: `AdsReadDeviceInfoRequest`, the online-version
  `AdsReadWriteRequest`, and the symbol-version `AdsReadRequest`.
- `onPing()` - the `AdsReadDeviceInfoRequest`.

All four are built with:

```java
getConfiguration().getTargetAmsNetId(), DefaultAmsPorts.RUNTIME_SYSTEM_01.getValue(),   // 851, hardcoded
getConfiguration().getSourceAmsNetId(), 800, ReturnCode.OK, getInvokeId());              // 800, hardcoded
```

whereas the symbol/datatype-table reads and all read/write requests in the same class use
`getConfiguration().getTargetAmsPort()` / `getSourceAmsPort()`.

The released `0.12.0` (and `0.13.x`) `AdsProtocolLogic.onConnect()`
hardcodes the same `RUNTIME_SYSTEM_01` / `800`, and the SPI-3 rewrite (`AdsTcpConnection`) carried it over.

### Steps to reproduce

1. A TwinCAT target with a PLC runtime on AMS port **852** (e.g. a second PLC project, or move the
   first project to 852), with **nothing** on 851.
2. Connect with `ads:tcp://<ip>?source-ams-net-id=...&source-ams-port=...&target-ams-net-id=...&target-ams-port=852`.
3. `getConnection(...)` fails during the handshake:
   `PlcConnectionException: Error during ADS connect handshake`
   -> `PlcRuntimeException: Unexpected response type: AdsErrorResponse (expected AdsReadDeviceInfoResponse)`.

A packet capture (Wireshark AMS dissector) confirms the on-connect `ReadDeviceInfo` is addressed to
AMS port **851** (source **800**) - not the configured 852 - and TwinCAT replies with AMS error
`0x00000006`. Sending the same `ReadDeviceInfo` to port 852 by hand (raw plain-ADS socket) succeeds and
returns the device info, so the target itself is fine.

### Expected behavior

The handshake should target `getConfiguration().getTargetAmsPort()` / `getSourceAmsPort()`, exactly
like the data path - so any runtime port works.

### Environment

- PLC4X: `0.14.0-SNAPSHOT` (develop); also present in released `0.12.0` / `0.13.x`.
- Driver: `plc4j-driver-ads` (Java).
- Target: Beckhoff TwinCAT `3.1.4026`.

### Proposed fix

Replace the four hardcoded `RUNTIME_SYSTEM_01.getValue()` / `800` with the configured
`getTargetAmsPort()` / `getSourceAmsPort()`. I have this fix on a branch and will open a PR referencing
this issue.

### Version

v0.14.0-SNAPSHOT

### Programming Languages

- [ ] plc4c
- [ ] plc4go
- [x] plc4j
- [ ] plc4net
- [ ] plc4py

### Protocols

- [ ] AB-Ethernet
- [x] ADS /AMS
- [ ] BACnet/IP
- [ ] C-Bus
- [ ] CANopen
- [ ] EtherNet/IP
- [ ] Firmata
- [ ] IEC-69870
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] Profinet
- [ ] S7
- [ ] S7-light

soldom:
Closes #2629.

### Problem

The ADS connect handshake (`AdsTcpConnection.doConnectHandshake()` and `onPing()`) sent its probe
requests — `ReadDeviceInfo`, online-version, symbol-version — to a hardcoded AMS port
`DefaultAmsPorts.RUNTIME_SYSTEM_01` (851) with a hardcoded source port `800`, ignoring the configured
`target-ams-port` / `source-ams-port`. Only the handshake was affected; the symbol/datatype-table
reads and the read/write path already use `getConfiguration().getTargetAmsPort()`.

Consequence: any PLC runtime not on 851 — a second PLC project (852), a non-default port, or parallel
runtimes — failed to connect during the handshake with ADS error `0x6` ("target port not found"),
even with `target-ams-port` set correctly. Full analysis and a packet capture are in #2629.

It may be intended to have the handshake port hardcoded to "System_01" (851), but the reason escapes me
and it does break the above mentioned PLC setup use cases.

### Change

In `plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/AdsTcpConnection.java`, the four
handshake/ping request builders now use the configured ports, matching the data path:

```diff
-    getConfiguration().getTargetAmsNetId(), DefaultAmsPorts.RUNTIME_SYSTEM_01.getValue(),
-    getConfiguration().getSourceAmsNetId(), 800, ReturnCode.OK, getInvokeId());
+    getConfiguration().getTargetAmsNetId(), getConfiguration().getTargetAmsPort(),
+    getConfiguration().getSourceAmsNetId(), getConfiguration().getSourceAmsPort(), ReturnCode.OK, getInvokeId());
```

No behavior change for the common case: `target-ams-port` already defaults to 851.

### Testing

Verified against a real Beckhoff TwinCAT `3.1.4026` IPC:

- PLC runtime on AMS port **852**, nothing on 851 — **before:** connect fails (`AdsErrorResponse`,
  AMS error 0x6); **after:** connects and reads a live symbol successfully.
- A packet capture (Wireshark AMS dissector) confirms the handshake `ReadDeviceInfo` is now sent to
  the configured `target-ams-port` (852) instead of 851.
- Regression check: with the PLC on 851 the connection still works as before.

### DIFF
--- a/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/AdsTcpConnection.java
+++ b/plc4j/drivers/ads/src/main/java/org/apache/plc4x/java/ads/AdsTcpConnection.java
@@ -190,8 +190,8 @@ public class AdsTcpConnection extends ConnectionBase<AdsConfiguration> {
     private CompletableFuture<Void> doConnectHandshake() {
         // ReadDeviceInfo
         AmsPacket readDeviceInfoRequest = new AdsReadDeviceInfoRequest(
-            getConfiguration().getTargetAmsNetId(), DefaultAmsPorts.RUNTIME_SYSTEM_01.getValue(),
-            getConfiguration().getSourceAmsNetId(), 800, ReturnCode.OK, getInvokeId());
+            getConfiguration().getTargetAmsNetId(), getConfiguration().getTargetAmsPort(),
+            getConfiguration().getSourceAmsNetId(), getConfiguration().getSourceAmsPort(), ReturnCode.OK, getInvokeId());
 
         return sendAmsRequest(readDeviceInfoRequest, AdsReadDeviceInfoResponse.class).thenCompose(readDeviceInfoResponse -> {
             if (readDeviceInfoResponse.getResult() != ReturnCode.OK) {
@@ -204,8 +204,8 @@ public class AdsTcpConnection extends ConnectionBase<AdsConfiguration> {
 
             // Read online version (sym-by-name "TwinCAT_SystemInfoVarList._AppInfo.OnlineChangeCnt")
             AmsPacket readOnlineVersionRequest = new AdsReadWriteRequest(
-                getConfiguration().getTargetAmsNetId(), DefaultAmsPorts.RUNTIME_SYSTEM_01.getValue(),
-                getConfiguration().getSourceAmsNetId(), 800, ReturnCode.OK, getInvokeId(),
+                getConfiguration().getTargetAmsNetId(), getConfiguration().getTargetAmsPort(),
+                getConfiguration().getSourceAmsNetId(), getConfiguration().getSourceAmsPort(), ReturnCode.OK, getInvokeId(),
                 ReservedIndexGroups.ADSIGRP_SYM_VALBYNAME.getValue(), 0L, 4L, null,
                 "TwinCAT_SystemInfoVarList._AppInfo.OnlineChangeCnt".getBytes(StandardCharsets.UTF_8));
             return sendAmsRequest(readOnlineVersionRequest, AdsReadWriteResponse.class);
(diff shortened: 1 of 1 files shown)


chrisdutz:
Just had a look and am a bit puzzled that nobody had noticed that yet. 
I just looked in the old version, if this was something I introduced with the refactoring, but this has always been there. 

Thanks for reporting and even providing a patch ... will merge it as soon as the build passes (However I see no reason why it should not pass) ... clicking merge now.

PR #2645, opened 2026-07-14
fix(plc4j/transports-tcp): close SocketChannel on failed bind/socket-option setup

### LINKED ISSUE #2644: [Bug]: S7 driver leaks Netty/protocol resources on repeated failed connect() attempts (orphaned resources, related to #2544)
### What happened?

We're running plc4j (S7 driver) inside a long-lived industrial gateway process on constrained embedded Linux hardware (512MB total RAM, JVM heap capped at `-Xmx128m`). When the remote S7 PLC becomes unreachable for an extended period (in our case ~3 hours, confirmed the PLC and network path were healthy — a purely transient outage), our application repeatedly calls `PlcDriverManager.getDefault().getConnectionManager().getConnection("s7://...")` on a backoff schedule (5-600s) to retry the connection. Each attempt times out after 5000ms without ever returning a `PlcConnection` instance to the caller.

Over the ~3 hour outage (roughly 17-18 failed connect attempts), the process's resident memory (RSS) climbed from its normal baseline to ~78% of the device's total RAM, at which point the Linux kernel OOM-killer killed the JVM (confirmed via `uptime` that the device itself never rebooted — only our process died and was restarted by our process supervisor). No Java `OutOfMemoryError` was ever logged, which is consistent with the growth happening outside the JVM heap (native/off-heap), invisible to heap-based OOM detection.

A thread dump taken shortly after the process restarted (before any new failures had occurred) already showed threads named `plc4x-tm-thread-*`, `plc4x-app-thread-*`, `plc4x-s7ha-thread-*`, `plc4x-evt-processor-*`, `plc4x-evt-dispatcher-*`, and `nioEventLoopGroup-*`.

This looks like the "Orphaned Resources on Failed Connection" scenario discussed (but not fixed) in #2544: since `connect()` fails before ever returning a connection object, `close()` is never invoked, and whatever `EmbeddedChannel` / `HashedWheelTimer` / protocol thread pool got allocated during that attempt is never released. #2544 was closed as "not planned" after the original reporter retracted their *thread*-leak claim for the normal-disconnect path, but the *failed-connect* orphaned-resource scenario was confirmed by maintainers on that same thread and, as far as we can tell, was never itself addressed. We're filing separately since our field data (RSS growth over time + thread dump + confirmed OOM-kill) is different evidence than what was available on #2544, and might help re-evaluate whether this specific path needs a fix.

**Steps to reproduce:**
1. Configure an S7 connection (`s7://<host>:102?remote-rack=0&remote-slot=1`) to an address that is reachable at the TCP level but never completes (or is simply firewalled/unreachable so `getConnection()` always times out).
2. Repeatedly call `getConnectionManager().getConnection(...)` on a timer (every 10-60s) for an extended period (hours), always against the same unreachable address.
3. Monitor process RSS and `/proc/<pid>/task/*/comm` thread names over time.

Happy to provide our thread-dump and RSS-over-time data if useful.

### Version

0.13.1

### Programming Languages

- [x] plc4j

### Protocols

- [x] S7

mkribs:
`TcpTransportInstance`'s constructor only assigned the opened `SocketChannel` to its field after `bind()`, socket options, and `connect()` all succeeded. If `bind()` or a socket-option call threw (e.g. a pinned local port still held by a previous failed attempt), the already-open channel was never closed - leaking one fd per failed attempt.

Note: `connect()` failures (timeout/refused) are **not** affected - `Socket.connect(SocketAddress, int)` already closes itself internally on failure (confirmed empirically). Only failures *before* `connect()` is reached (bind, socket options) leak.

**Fix:** hold the channel in a local variable, only promote it to the `socketChannel` field once every setup step succeeds; explicitly close it in the constructor's catch block otherwise.

**Testing:** added `testConstructor_failedBind_doesNotLeakFileDescriptors`, which deterministically forces a `BindException` (occupies a local port, configures the transport to bind to that same port) and asserts the open-file-descriptor count stays flat across repeated failures via `UnixOperatingSystemMXBean`. Verified the test fails without this fix (fd count grows ~1 per attempt) and passes with it, on Linux.

closes #2644

### DIFF
--- a/plc4j/transports/tcp/src/main/java/org/apache/plc4x/java/transport/tcp/TcpTransportInstance.java
+++ b/plc4j/transports/tcp/src/main/java/org/apache/plc4x/java/transport/tcp/TcpTransportInstance.java
@@ -119,18 +119,13 @@ public class TcpTransportInstance extends BaseTransportInstance<TcpTransportConf
                 remoteAddress.getHostName(), remoteAddress.getPort(),
                 getLocalAddress().getHostName(), getLocalAddress().getPort()));
 
-            // Start the per-connection read loop on a virtual thread (Java 21+) LAST, so an
-            // unchecked throw from the logging/audit above cannot leak an already-running thread
-            // (the catch only handles IOException and does not stop the read loop).
+            // Start the per-connection read loop on a virtual thread (Java 21+) LAST, so a
+            // throw from the logging/audit above cannot leak an already-running thread.
             this.readThread = Thread.ofVirtual()
                 .name("TCP-Read-" + remoteAddress.getHostName() + ":" + remoteAddress.getPort())
                 .start(this::runReadLoop);
-        } catch (IOException e) {
-            String errorMsg = String.format("Failed to connect to %s:%d - %s",
-                remoteAddress.getHostName(), remoteAddress.getPort(), e.getMessage());
-            LOGGER.error(errorMsg, e);
-            // errorMsg already embeds e.getMessage(); a single audit event avoids a duplicate.
-            auditLog.write(AuditLogEventType.ERROR, "Error in constructor: " + errorMsg);
+        } catch (Exception e) {
+            // Close before any other side effect (logging/audit), so a failure there can't skip cleanup.
             if (channel != null) {
                 try {
                     channel.close();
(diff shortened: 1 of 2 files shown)


sruehl (changes requested):
see above comment

PR #2673, opened 2026-08-05
fix(plc4j/eip): fix ConnectionManager

andvasp:
The field connectionSerialNumber is out of range to establish connection
Add missing encoding for unsigned integers to reading tags

### DIFF
--- a/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
+++ b/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
@@ -29,10 +29,8 @@ import org.apache.plc4x.java.eip.base.configuration.EIPConfiguration;
 import org.apache.plc4x.java.eip.base.tag.EipTag;
 import org.apache.plc4x.java.eip.base.tag.EipTagHandler;
 import org.apache.plc4x.java.eip.readwrite.*;
-import org.apache.plc4x.java.spi.buffers.api.WithOption;
 import org.apache.plc4x.java.spi.buffers.api.exceptions.BufferException;
 import org.apache.plc4x.java.spi.buffers.bytebased.ReadBufferByteBased;
-import org.apache.plc4x.java.spi.buffers.bytebased.WithByteBasedOption;
 import org.apache.plc4x.java.spi.buffers.bytebased.WriteBufferByteBased;
 import org.apache.plc4x.java.utils.subscriptionemulation.PollingSubscriptionConnectionBase;
 import org.apache.plc4x.java.spi.drivers.exceptions.MessageCodecException;
@@ -93,7 +91,7 @@ public class EipTcpConnection extends PollingSubscriptionConnectionBase<EIPConfi
     private final NullAddressItem nullAddressItem = new NullAddressItem();
     private final List<PathSegment> routingAddress = new ArrayList<>();
     private short connectionPathSize = 0;
-    private final int connectionSerialNumber = ThreadLocalRandom.current().nextInt();
+    private final int connectionSerialNumber = ThreadLocalRandom.current().nextInt(1, 0xFFFF);
 
     public EipTcpConnection(EIPConfiguration configuration, TransportInstance<?> transportInstance, AuditLog auditLog) {
         this(configuration, transportInstance, auditLog, configuration.isBigEndian());
@@ -749,7 +747,7 @@ public class EipTcpConnection extends PollingSubscriptionConnectionBase<EIPConfi
     // Encoders / decoders
     ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-    private static byte[] toAnsi(String tag) throws BufferException {
+    private byte[] toAnsi(String tag) throws BufferException {
         Pattern resourcePattern = Pattern.compile("([.\\[\\]])*([A-Za-z_0-9]+)");
         Matcher matcher = resourcePattern.matcher(tag);
         List<PathSegment> segments = new LinkedList<>();
@@ -767,12 +765,7 @@ public class EipTcpConnection extends PollingSubscriptionConnectionBase<EIPConfi
             segments.add(newSegment);
             lengthBytes += newSegment.getLengthInBytes();
         }
-        WriteBufferByteBased buffer = new WriteBufferByteBased(new byte[lengthBytes],
-            WithByteBasedOption.WithByteOrder("LITTLE_ENDIAN"),
-            WithOption.WithUnsignedIntegerEncoding("unsigned-binary"),
-            WithOption.WithSignedIntegerEncoding("twos-complement"),
-            WithOption.WithFloatEncoding("IEEE754"),
-            WithOption.WithStringEncoding("UTF8"));
+        WriteBufferByteBased buffer = messageCodec.createWriteBuffer(lengthBytes);
         for (PathSegment segment : segments) {
             segment.serialize(buffer);
         }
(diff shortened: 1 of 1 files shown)


plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -767,12 +765,7 @@ private static byte[] toAnsi(String tag) throws BufferException {
             segments.add(newSegment);
             lengthBytes += newSegment.getLengthInBytes();
         }
-        WriteBufferByteBased buffer = new WriteBufferByteBased(new byte[lengthBytes],
-            WithByteBasedOption.WithByteOrder("LITTLE_ENDIAN"),
-            WithOption.WithUnsignedIntegerEncoding("unsigned-binary"),
-            WithOption.WithSignedIntegerEncoding("twos-complement"),
-            WithOption.WithFloatEncoding("IEEE754"),
```
chrisdutz, line 774: I'm a bit surprized this would work ... if I look in mspec, the type PathSegment doesn't define the different encodings. So they should be null if you remove them all. It should require setting the unsigned integer and the string encodings. Also would you be changing the endianess from little-endian to big-endian.

chrisdutz: If you look at EipTcpMessageCodec this should take care of creating the write buffer with a call to "createWriteBuffer" ... this should also automatically handle the case of a big-endian and little-endian device. So I think your system might be a big-endian one and for that reason this part of the code probably failed and removing the byte order should make it default to big-endian.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -795,8 +788,7 @@ private PlcReadResponse decodeReadResponse(CipService p, PlcReadRequest readRequ
             List<CipService> arr = new ArrayList<>(nb);
             try {
                 byte[] servicesData = responses.getServicesData();
-                ReadBufferByteBased read = new ReadBufferByteBased(servicesData,
-                    WithByteBasedOption.WithByteOrder("LITTLE_ENDIAN"));
+                ReadBufferByteBased read = messageCodec.createReadBuffer(servicesData);
```
chrisdutz, line 791: As mentioned in the other change ... I assume the problem you were having was that your system was a big-endian one and all of mine were little-endian ones. This change looks a lot better than the one in lines 774++

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -93,7 +91,7 @@ public class EipTcpConnection extends PollingSubscriptionConnectionBase<EIPConfi
     private final NullAddressItem nullAddressItem = new NullAddressItem();
     private final List<PathSegment> routingAddress = new ArrayList<>();
     private short connectionPathSize = 0;
-    private final int connectionSerialNumber = ThreadLocalRandom.current().nextInt();
+    private final int connectionSerialNumber = ThreadLocalRandom.current().nextInt(1, 0xFFFF);
```
chrisdutz, line 94: Great job spotting this ... indeed the field is defined as unsigned int 16 bit ... and for that 0xFFFF is indeed the max value.

chrisdutz (changes requested):

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: Yes. I am using EipTcpMessageCodec to create the buffer. Look at added line below. 

There are other places on EipTcpConnection that is using LITTLE_ENDIAN (encodeValue and parsePlcValue). Shold be used the byte order from configuration?

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: I tested here just changing "LITTLE_ENDIAN" to "BIG_ENDIAN" and still raising the same error: "No encoding defined for unsigned integer values" when using connection manager.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: > I'm a bit surprized this would work ... if I look in mspec, the type PathSegment doesn't define the different encodings. So they should be null if you remove them all. It should require setting the unsigned integer and the string encodings. Also would you be changing the endianess from little-endian to big-endian.

Now that you mentioned the point about byte order, I am surprised too why it was working because it looks like it is mixing the byte order. Let me summarise my scenario:

- I did not change the configuration of the byte order. So I am using big endian.
- I can read the value of most of the tags.
- There are some tags that I cannot, even on same device (addressing this issue later) but I did not check them after the modification of this PR.
- I did not have the opportunity to test the write value but I can and I plan to do.

chrisdutz: I think my system used Little endian and my code was obviously bad. So making it dynamic like in the other cases is the way forward ... if you find more places, where I created the buffer manually, I think it would be good to allways use the transportCodec functions for generating them.

chrisdutz: Also just noticed that you already had it right ... I shouldn't play with crazy protocols to deep in the night ;-) Sorry for the confusion ... will approve it right away.

chrisdutz (approved):
Was already Ok .... no Idea what I saw there last time ;-)

chrisdutz:
Argh!!!! ... stupid signatures :-( (I think we're stalled here till that "feature" is turned off again :-( )

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: >  if you find more places, where I created the buffer manually, I think it would be good to allways use the 
> transportCodec functions for generating them.

I found places using  LITTLE_ENDIAN not creating the buffer but ByteBuffer show below. I do not know if would be a problem.

What do you think?

```java
private byte[] encodeValue(PlcValue value, CIPDataTypeCode type) {
	ByteBuffer buffer = ByteBuffer.allocate(type.getSize()).order(ByteOrder.LITTLE_ENDIAN);

```
```java
private PlcValue parsePlcValue(EipTag tag, byte[] rawData, CIPDataTypeCode type) {
	final int STRING_LEN_OFFSET = 2;
	final int STRING_DATA_OFFSET = 6;
	ByteBuffer data = ByteBuffer.wrap(rawData).order(ByteOrder.LITTLE_ENDIAN);

```

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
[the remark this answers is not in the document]
chrisdutz: Think I'm just going to merge this and fix this myself after merging ... don't want to go another set of rounds for this tiny change.

Issue #2135, opened 2025-06-06
[Bug]: Integration Issues with Rockwell ControlLogix PLC


korni128:
## Description
We are currently working on integrating a Rockwell PLC using the PLC4J library with the Ethernet/IP Protocol. While our implementation successfully communicates with a Rockwell CompactLogix PLC, we are encountering issues when attempting to connect to a Rockwell ControlLogix PLC.

## Observations
During our analysis with Wireshark, we observed the following:

- The initial `GetAttributesAll` call to the ControlLogix PLC fails, returning the error: 'General Status: Service not supported (0x08).'
- In contrast, the same call works without issues on the CompactLogix PLC, where we receive responses with attributes for: 'Class: Message Router (0x0002)' and 'Class: Connection Manager (0x0006)'

## Current Understanding
If the `GetAttributesAll` call fails, then the `useConnectionManager` and `useMessageRouter` boolean values are set to `false` in the `EipProtocolLogic` class. As a result, PLC4X is attempting to perform single requests.

### Issues Identified
- Sending single requests sequentially is causing issues related to the PLC4X library.
- Sending single requests when `useMessageRouter` and `useConnectionManager` are not active has also been problematic in the past in the PLC4X library, based on my understanding of the code. Therfore I would prefer the way via Message Router / Connection Manager.

## Idea
- The `GetAttributeSingle` method may serve as a potential workaround instead of using GetAttributesAll, in order to get the attributes 'Class: Message Router (0x0002)' and 'Class: Connection Manager (0x0006)' one after the other. However, if the Message Router and Connection Manager are not functioning correctly, implementing this workaround may not solve the issue and we have also not tried to use GetAttributeSingle (just seen it in some documentation).

## Questions
1. Does anyone know why the `GetAttributesAll` call may be failing with the ControlLogix PLC? 
2. Any ideas on potential workarounds or solutions would be greatly appreciated.

Thank you 

### Version

v0.12.0

### Programming Languages

- [ ] plc4c
- [ ] plc4go
- [x] plc4j
- [ ] plc4net
- [ ] plc4py

### Protocols

- [ ] AB-Ethernet
- [ ] ADS /AMS
- [ ] BACnet/IP
- [ ] C-Bus
- [ ] CANopen
- [x] EtherNet/IP
- [ ] Firmata
- [ ] IEC-69870
- [ ] KNXnet/IP
- [ ] Modbus
- [ ] OPC-UA
- [ ] Profinet
- [ ] S7
[...]

PR #2674, opened 2026-08-05
feat(plc4j/eip):  Detect Connection Manager and Message using "Get Attribute Single" 

andvasp:
Hi @chrisdutz ,

Implement "Get Attribute Single" on EIP protocol for cases where "Get Attribute All" is not supported by the device to address  #2135.


plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -249,16 +249,13 @@ private CompletableFuture<Void> probeAttributes() {
...
-            if (!(dataItem.getService() instanceof GetAttributeAllResponse gar)) {
-                return;
+        return sendRequest(eipWrapper).thenCompose(response -> {
+            if (!(response instanceof CipRRData rr) || rr.getStatus() != CIPStatus.Success.getValue() ||
+                !(rr.getTypeIds().get(1) instanceof UnConnectedDataItem di && di.getService() instanceof GetAttributeAllResponse gar)) {
+                return CompletableFuture.completedFuture(null);
             }
             if (gar.getStatus() == CIPStatus.ServiceNotSupported.getValue()) {
-                return;
+                return checkAttributesSingle();
```
chrisdutz: Admittedly I find this particular part of code quite hard to read ... Could you please simplify this a bit? 

Admittedly I'm a big fan of the "di.getService() instanceof GetAttributeAllResponse gar" notation saving myself the explicit cast, but I am super unhappy with the decision of the Java group in a negated form to make the variable available outside the if statement (Which you are using) ... it's just challenging from a maintenance perspective.

chrisdutz: Also leaving a few comments here to what's happeing would be great. I know I didn't set a good example but I'm trying my best to leave more comments for my fellow maintainers if I think something's tricky to understand. I guess the problematic you're trying to solve would qualify for such a comment.

chrisdutz (changes requested):

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: > Admittedly I find this particular part of code quite hard to read ... Could you please simplify this a bit?
> 
> Admittedly I'm a big fan of the "di.getService() instanceof GetAttributeAllResponse gar" notation saving myself the explicit cast, but I am super unhappy with the decision of the Java group in a negated form to make the variable available outside the if statement (Which you are using) ... it's just challenging from a maintenance perspective.

Here I see 3 options, where I prefer the sequential other. What do you think?

Option 1:
if (!(response instanceof CipRRData rr) || rr.getStatus() != CIPStatus.Success.getValue() ||
	!(rr.getTypeIds().get(1) instanceof UnConnectedDataItem di) ||
	!(di.getService() instanceof GetAttributeAllResponse gar)) {
	return CompletableFuture.completedFuture(null);
}

Option 2:
if (!(response instanceof CipRRData rr) || rr.getStatus() != CIPStatus.Success.getValue()) {
	return CompletableFuture.completedFuture(null);
}
if (!(rr.getTypeIds().get(1) instanceof UnConnectedDataItem di) ||
	!(di.getService() instanceof GetAttributeAllResponse gar)) {
	return CompletableFuture.completedFuture(null);
}

Option 3:
if (!(response instanceof CipRRData rr) || rr.getStatus() != CIPStatus.Success.getValue()) {
	return CompletableFuture.completedFuture(null);
}
UnConnectedDataItem dataItem = (UnConnectedDataItem) rr.getTypeIds().get(1);
if (!(dataItem.getService() instanceof GetAttributeAllResponse gar)) {
	return CompletableFuture.completedFuture(null);
}

chrisdutz: I'm more trying to wrap my head around what the code should do ... 

If it's (not a CipRRData) or (it is and it's status is not success) or (it is and it's first type id is an UnConnectedDataItem) or (it is, it's first type id is a UnConnectedDataItem's service is a GetAttributeAllResponse ....

So much negation ... wouldn't it be an alternative to focus what we expect it to be?
Something like this?

```
if (response instanceof CipRRData rr
            && rr.getStatus() == CIPStatus.Success.getValue()
            && rr.getTypeIds().size() > 1
            && rr.getTypeIds().get(1) instanceof UnConnectedDataItem di
            && serviceType.isInstance(di.getService())) {
        return serviceType.cast(di.getService());
    }
    return null;
```

That I would instantly understand ;-)

andvasp: Hi Chris!

I push a new version. Please check if it is better now.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -278,11 +278,21 @@ private CompletableFuture<Void> probeAttributes() {
         });
     }
 
-    private CompletableFuture<Void> checkAttributesSingle() {
+    private CipService getCipService(EipPacket response) {
```
andvasp: I created this method to encapsulate the logic for get the CipService. I believe it could also be applied elsewhere, even though the logic isn't exactly the same.

Let me know what you think.

chrisdutz: A simple getCipService would make me expect that it simply gets the CIP service ... here the method is actually extracting something if a very specific data-case is present or doesn't do anything if that's not the case.

Are we using this or could we use this in different places? 

In the past we used a lot of "return null" methods and are more and more trying to use Optionals in Java ... I tink renaming it to something that indicates the fact that it's not just blindly accessing something, possibly something like: 

```
Optional<CipService> extractCipService(EipPacket response) 
```

Might not hide this detail?

chrisdutz (changes requested):

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
andvasp: > A simple getCipService would make me expect that it simply gets the CIP service ... here the method is actually extracting something if a very specific data-case is present or doesn't do anything if that's not the case.

Yes. extractCipService is a better name. Actually, I thought about using it when thinking about possible solutions but when implementing it I forget to use it.
 
> Are we using this or could we use this in different places?

I am using this method at 2 places. I found other places that could use it but as they do not have exactly the same logic, I prefer to not change them. But I believe would be good to use. Let me know what you think.

> In the past we used a lot of "return null" methods and are more and more trying to use Optionals in Java ... I tink renaming it to something that indicates the fact that it's not just blindly accessing something, possibly something like:
> 
> ```
> Optional<CipService> extractCipService(EipPacket response) 
> ```

I considered using Optional, but in this case I found it more verbose and as this method is used just internally in this class and with Pattern Matching, I prefer to not use Optional here. See the examples below and tell me what do you think.

```java
if (service.isPresent() && service.get() instanceof GetAttributeSingleResponse gsr) // I do not like to use isPresent() and get() together but I believe is less verbose.
```

I considered create the method below but again I think is more verbose and it with extra comparations at Class methods.
```java
<T> Optional<T> extractCipService(EipPacket response, Class<T> type)
```

chrisdutz: What do you think about this? As I mentioned ... I'm not very fond of these cast variables outside of the code block that creates it? (And yes ... I know you didn't introduce this pattern ;-)

```
return switch (getCipService(response)) {
        case GetAttributeAllResponse gar
                when gar.getStatus() == CIPStatus.ServiceNotSupported.getValue() ->
            probeAttributesUsingSingleAttributeRequest();
        case GetAttributeAllResponse gar -> {
            recordSupportedClasses(gar.getAttributes());
            yield CompletableFuture.completedFuture(null);
        }
        case null, default -> CompletableFuture.completedFuture(null);
    };
```

andvasp: I liked the suggestion! I will do it.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -280,6 +278,48 @@ private CompletableFuture<Void> probeAttributes() {
...
+            useMessageRouter = hasSupport;
+        });
+    }
+
+    private CompletableFuture<Boolean> checkAttributeSupport(CIPClassID classId) {
+
+        UnConnectedDataItem exchange = new UnConnectedDataItem(new GetAttributeSingleRequest(
+            new LogicalSegment(new ClassID((byte) 0, (short) classId.getValue())),
+            new LogicalSegment(new InstanceID((byte) 0, (short) 0)), // Class level discovery
+            new LogicalSegment(new AttributeID((byte) 0, (short) 1))) // Attribute ID 1: Revision
```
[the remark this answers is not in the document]
andvasp: I was using instanceId 1 at previous commit. I changed it after search and find out that 0 is used to discover. 
Looks like the PLC LOGIX5573 supports message router and connection manager. When I test it using instanceId = 0 I receive the information that it support but does not when using instance = 1. So I believe instanceId = 0 is correct.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -280,6 +278,48 @@ private CompletableFuture<Void> probeAttributes() {
...
+    private CompletableFuture<Void> probeAttributesUsingSingleAttributeRequest() {
+        LOGGER.debug("Checking MessageRouter and ConnectionManager using GetAttributeSingle");
+
+        return checkAttributeSupport(CIPClassID.ConnectionManager).thenCompose(hasSupport -> {
+            useConnectionManager = hasSupport;
+            return checkAttributeSupport(CIPClassID.MessageRouter);
+        }).thenAccept(hasSupport -> {
+            useMessageRouter = hasSupport;
+        });
+    }
```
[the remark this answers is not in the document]
andvasp: If we can check that the connection manager is available and an error happens while checking for message router support, I believe we should keep the connection manager as available and not fallback to unconnected code path. This way we would just update the comments on the exceptionally block.

What do you think @chrisdutz ?

protocols/eip/src/main/resources/protocols/eip/eip.mspec
```diff
@@ -157,10 +157,17 @@
...
+            [simple     PathSegment         instanceSegment                                                            ]
+            [simple     PathSegment         attributeSegment                                                           ]
         ]
         ['0x0E','true'  GetAttributeSingleResponse
-            // TODO: Implement
+            [reserved   uint    8           '0x00'                                                                     ]
+            [simple     uint    8           status                                                                     ]
+            [simple     uint    8           extStatusSize                                                              ]
+            [array      uint    8           extStatus count 'extStatusSize'                                            ]
+            [array      byte   servicesData count 'serviceLen - 4 - extStatusSize'                                     ]
```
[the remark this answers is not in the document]
andvasp: Yes. Now I'm curious.

The "header" message specification layout is the same in both messages (Reply Service, Reserved, Status, Additional Status Size, and Additional Status).

I don't know much about *mspec*, but it looks to me like the `extStatusSize` field is missing from `GetAttributeAllResponse`. Am I doing something wrong with `GetAttributeSingleResponse`? Why is there this difference? 

@chrisdutz, could you clarify this?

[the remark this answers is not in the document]
chrisdutz: Hehe ... well ... I could immagine, that we simply got the others wrong ;-)
I would need to check if possibly all others we only had extStatus=0 ... in that case our other cases would have worked. I'll let Claude do some research ... however my gut-feeling tells me that I would expect your version to be right and ours to be wrong.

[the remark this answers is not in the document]
chrisdutz: Yup ... you were right ... guess we need to update our mspec ;-)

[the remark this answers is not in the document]
andvasp: ok. Thank you!

andvasp:
> ## Pull request overview
> Copilot reviewed 3 out of 7 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (3)
> **protocols/eip/src/main/resources/protocols/eip/eip.mspec:170**
> 
> * `extStatusSize` is the number of 16-bit additional-status words, but this models each entry as 8 bits and subtracts only one byte per entry. Any response with a nonzero additional status will therefore split the status word and expose its second byte as `servicesData`; serialization also emits an invalid packet. Define `extStatus` as `uint 16`, subtract `2 * extStatusSize`, regenerate the generated models, and cover a nonzero additional-status response.
> 
> ```
>             [array      uint    8           extStatus count 'extStatusSize'                                            ]
>             [array      byte   servicesData count 'serviceLen - 4 - extStatusSize'                                     ]
> ```
Would be just necessary change uint  8 to 16 like below?
```
             [array      uint    16           extStatus count 'extStatusSize'                                            ]
             [array      byte   servicesData count 'serviceLen - 4 -  2 * extStatusSize'                                     ]

```

andvasp:
Hi @chrisdutz.

Reading through the code, I understand that we use Get_Attribute_All to check if the PLC supports the Connection Manager (CM) and Message Router (MR).

However, while investigating an issue, I noticed that when we use Get_Attribute_All, we must specify which class attributes we are requesting. Currently, the code is only searching for attributes of the MR (using the value 2 in the snippet below). Because of this, when I received a ServiceNotSupported status, it actually meant the MR was not supported, rather than the entire connection probe failing.

```java
private CompletableFuture<Void> probeAttributes() {
    PathSegment classSegment = new LogicalSegment(new ClassID((byte) 0, (short) 2));
```

Which class has its attributes, for example:
Identity (Vendor ID, Device Type, Product Code,..) 
Connection Manager (Revision, Max Instance, Number of Instances)

If you agree, I will submit a modification accordingly.

chrisdutz:
Please don't treat that I was super knowledgable about all the variants of EIP when I built that driver ... if you find better ways to do so, please give it a try ... and if you write code comments explaining the rationale behind it, that would be 100 extra bonus points ;-) (That's also a reason I like code comments with protocol details ... it leaves information behind for the rest to learn from)

andvasp:
Hi @chrisdutz ,

I have not updated this PR because I am facing problems running the application on Intellij with maven 4.

Would it be a problem to rebase develop on this PR?

chrisdutz:
Öhm ... Rebasing "develop" is considered really bad. Rebasing this PR on develop I'd say ok ... I usually rebase my local develop on the remote develop so commit history becomes more linear ... If you're working on a branch and you're sure nobody else has checked that out, pushing a rebased branch could be ok ... but I'd reccommend merging.

In your case, merging in changes from plc4x/develop is probably the best option.

chrisdutz:
> Hi @chrisdutz ,
> 
> I have not updated this PR because I am facing problems running the application on Intellij with maven 4.
> 
> Would it be a problem to rebase develop on this PR?

If you tell me what problems you are having, I might be able to help you as I'm using IntelliJ myself.

andvasp:
> If you tell me what problems you are having, I might be able to help you as I'm using IntelliJ myself.

Thank you Chris for the help.

I am trying to debug the main method org.apache.plc4x.java.eip.base.manual.ManualEipTest at plc4j-driver-eip.

It stopped giving this possibility after I updated my Intellij. I was using "Bundled (Maven 3)" at the "Maven home path" then I had to change to "Use Maven wrapper" to package the project.

When I open the ManualEipTest it shows:
"Java file is located outside of the module source root, so it... "
It shows the right arrow green to execute the main method but when I click on it appear "Nothing here"

My scenario:
IntelliJ IDEA 2026.2.1
Windows
Maven home path: Use Maven wrapper

andvasp:
> If you tell me what problems you are having, I might be able to help you as I'm using IntelliJ myself.

I could make it wotks now. 

Thank you to offer a help.

chrisdutz:
HI @andvasp ... what's the status on this? You able to continue? You waiting for something from my side?

Just asking as this is expected to address the last open issue we have in GH.

Will be focussing on some other polishing over the next few days, but would love to resolve this one as quickly as possible so we can start the release process.

andvasp:
Hi @chrisdutz .

It is ready for review. I request a review from github. Sorry, I do not know the right mechanism to notify.

Some points:

- Now I am using GET_ATTRIBUTE_SINGLE directly instead of GET_ATTRIBUTE_ALL since the attributes itself does not inform the supported classes. Could be used but doing a request for each class like we are using on GET_ATTRIBUTE_SINGLE.

- GET_ATTRIBUTE_ALL continues to be used on Identity class just to not lose this method because I believe it can be useful in future.

- I did not change GET_ATTRIBUTE_ALL mspec to include additional status size since it will not impact the current implementation. Would be good to create an issue to change this for all messages that have additional status size.

chrisdutz (approved):

andvasp:
Thank you @chrisdutz for your approval.

Sorry to borrow you... But Did I do something wrong on requesting this PR? Because it has not been merged into develop.

I believe I create it to merged not at develop. What is the best way to fix it?

chrisdutz:
No ... not at all. There was a merge conflict that needed resolving before I could even merge it ... I'm working on it taking it a bit further (You noticed our wrong extStatus fields and I'm just taking the opporthnity to clean things up ... as soon as all is finished, I'll merge it back.

chrisdutz:
Ok .... and I've merged things back to develop ... I assume you didn't run the build for go, c, python ... I needed to do a bit more than simply remove merge conflicts ... I also pulled in some changes that I had in my other repo.

PR #2707, opened 2026-08-21
chore: Updated the THREAT-MODEL.md to exclude not-supported drivers.

chrisdutz:
In order to reduce the number of false positives.


sruehl:
I'll check if I can fix that before 1.0 so we can leave those in the reactor at least

sruehl (changes requested):
maybe only doc for now and leave the reactor as is

sruehl:
and on the other hand it is fine to get those reports so we can fix them. Dunno if we merge this then we are blind there, right?

sruehl:
@chrisdutz guess this became obsolete

chrisdutz:
Yup

PR #2711, opened 2026-08-24
feat(plc4j/eip): Add support for DWORD data type

andvasp:
Add support for DWORD data type.

Would it be good to provide LWORD data type too?

I know that documentation should be updated too but in this case how to do it since the support here is just for plc4j?

### DIFF
--- a/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
+++ b/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
@@ -955,91 +955,66 @@ public class EipTcpConnection extends PollingSubscriptionConnectionBase<EIPConfi
     /** Whether the type is stored as a flat sequence of equally sized elements. */
     private static boolean isFixedSize(CIPDataTypeCode type) {
         return switch (type) {
-            case SINT, INT, DINT, LINT, REAL, LREAL, BOOL -> true;
+            case SINT, INT, DINT, DWORD, LINT, REAL, LREAL, BOOL -> true;
             default -> false;
         };
     }
 
     // Package-private and static so the decoding can be tested without a connection.
     static PlcValue parsePlcValue(EipTag tag, byte[] rawData, CIPDataTypeCode type) {
-        final int STRING_LEN_OFFSET = 2;
-        final int STRING_DATA_OFFSET = 6;
         ByteBuffer data = ByteBuffer.wrap(rawData).order(ByteOrder.LITTLE_ENDIAN);
         int nb = elementCount(tag);
-        if (nb > 1) {
-            // Never read past what the device actually sent - a short reply must not blow up
-            // the whole response with an IndexOutOfBoundsException. Only the fixed-size types
-            // below are laid out element by element; STRING/STRUCTURED carry their own length.
-            int elementSize = isFixedSize(type) ? type.getSize() : 0;
-            if (elementSize > 0 && rawData.length < nb * elementSize) {
-                LOGGER.warn("Device returned {} bytes for tag '{}', expected {} for {} elements of {}.",
-                    rawData.length, tag.getTag(), nb * elementSize, nb, type);
+        if (nb == 1) {
+            return extractPlcValue(rawData, type, data, 0);
+        }
+        // Never read past what the device actually sent - a short reply must not blow up
+        // the whole response with an IndexOutOfBoundsException. Only the fixed-size types
+        // below are laid out element by element; STRING/STRUCTURED carry their own length.
+        int elementSize = isFixedSize(type) ? type.getSize() : 0;
+        if (elementSize > 0 && rawData.length < nb * elementSize) {
+            LOGGER.warn("Device returned {} bytes for tag '{}', expected {} for {} elements of {}.",
+                rawData.length, tag.getTag(), nb * elementSize, nb, type);
+            return null;
+        }
+        List<PlcValue> list = new ArrayList<>();
+        int index = 0;
+        for (int i = 0; i < nb; i++) {
+            PlcValue plcValue = extractPlcValue(rawData, type, data, index);
+            if (plcValue == null) {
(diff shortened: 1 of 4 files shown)


plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -1013,91 +955,66 @@ private static int elementCount(EipTag tag) {
...
+            if (plcValue instanceof PlcSTRING) {
+                return new PlcList(list);
             }
-            return new PlcList(list);
         }
+        return new PlcList(list);
+    }
+
+    private static PlcValue extractPlcValue(byte[] rawData, CIPDataTypeCode type, ByteBuffer data, int index) {
+        final int STRING_LEN_OFFSET = 2;
```
andvasp, line 996: what about make these two variables (STRING_LEN_OFFSET and STRING_DATA_OFFSET) static?

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -1013,91 +955,66 @@ private static int elementCount(EipTag tag) {
...
-                        if (structuredType == CIPStructTypeCode.STRING.getValue()) {
-                            list.add(new PlcSTRING(new String(rawData, STRING_DATA_OFFSET, structuredLen, StandardCharsets.UTF_8)));
-                        }
-                        return list.isEmpty() ? null : new PlcList(list);
-                    default:
-                        return null;
-                }
+            list.add(plcValue);
+            index += type.getSize();
+            if (plcValue instanceof PlcSTRING) {
```
andvasp, line 988: Here I am doing this "if (plcValue instanceof PlcSTRING)" to have the same previous behavior but I believe it is not necessary.

andvasp:
Hi @chrisdutz and @sruehl .

If possible please take a look at this PR before release 1.0.

website/asciidoc/modules/users/pages/protocols/eip.adoc
```diff
@@ -70,6 +70,7 @@ These are the data types the driver can encode and decode:
 |===
 |To store a|Use this data type
 |Bit|BOOL
+|Bit array|DWORD (32-bit boolean array)
```
chrisdutz, line 73: Well ... wouldn't BYTE be used for storing 8-bit boolean arrays, WORD for 16 bit and LWORD for 64 bit arrays? Why only mention DWORD?

protocols/eip/src/main/resources/protocols/eip/eip.mspec
```diff
@@ -478,7 +478,7 @@
     ['0X00D1'   BYTE            ['1']]
     ['0X00D2'   WORD            ['2']]
     ['0X00D3'   DWORD           ['4']]
-    ['0X00D3'   LWORD           ['8']]
+    ['0X00D4'   LWORD           ['8']]
```
chrisdutz, line 481: Good find.

chrisdutz:
Admittedly some of this PR look legit, just don't understand the entire rewrite of the parsePlcValue method ... would love to hear why you did that. Seems to handle more than just adding DWORD support.

chrisdutz (changes requested):

chrisdutz:
And yes ... while at it .. of course should we then also add LWORD support.

andvasp:
> Admittedly some of this PR look legit, just don't understand the entire rewrite of the parsePlcValue method ... would love to hear why you did that. Seems to handle more than just adding DWORD support.

Hi @chrisdutz,

My apologies for changing the structure of parsePlcValue alongside the new feature!

The main goal was to remove duplicate switch statements so new data types like DWORD (and (and potentially LWORD) could be added in a single place rather than 2 places. I also inverted the if check to improve readability, which caused some line formatting/whitespace changes.

If you prefer to keep this PR strictly focused on adding DWORD support, I’d be happy to revert parsePlcValue to its original structure and submit the refactoring as a separate PR!

website/asciidoc/modules/users/pages/protocols/eip.adoc
andvasp: Good question, it makes sense but this is something that I am not sure about. This text I got from the previous version of this file which matches the behavior of my tests.

andvasp:
> And yes ... while at it .. of course should we then also add LWORD support.

Yes! Definitely I thought about doing it. Just let me know if you want me to revert the refactoring done here.

plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/EipTcpConnection.java
```diff
@@ -1013,91 +955,66 @@ private static int elementCount(EipTag tag) {
...
-                return new PlcSINT(data.get(0));
+                return new PlcSINT(data.get(index));
             case INT:
-                return new PlcINT(data.getShort(0));
+                return new PlcINT(data.getShort(index));
             case DINT:
-                return new PlcDINT(data.getInt(0));
+                return new PlcDINT(data.getInt(index));
+            case DWORD:
+                return new PlcDWORD(data.getInt(index));
```
chrisdutz: A DWORD is an unsigned value, so instead of using the Java 32 bit int value, we should use a long value.

chrisdutz:
So ... while reviewing it turns out there were more code collisions in the mspec constants and even with LWORD support the driver wouldn't have been able to decode BYTE and WORD ... so I'm whipping up my own fix for this as I really want to start the release process today.

chrisdutz:
Turns out also the unsigned integer types were missing too ...

chrisdutz:
Ok ... so I've added DWORD, LWORD, USINT, UINT, ULINT and UDINT to the driver. ... so I guess we can close this pr ... thanks for spotting some of the things here and for your work.

PR #2726, opened 2026-08-31
feat(plc4j/eip): Compute elementsNb once instead of doing it on every call

andvasp:
Hi @chrisdutz!

This PR refactors EipTag to compute elementsNb once during initialization rather than recalculating it on every call to getElementNb().

Please let me know if you agree with this approach or if you have any feedback!

Thanks,
Anderson

### DIFF
--- a/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/tag/EipTag.java
+++ b/plc4j/drivers/eip/src/main/java/org/apache/plc4x/java/eip/base/tag/EipTag.java
@@ -80,6 +80,7 @@ public class EipTag implements PlcTag, Serializable {
 
     private final String tag;
     private final CIPDataTypeCode type;
+    private final int elementsNb;
     /**
      * What the address selects. Drives the CIP path and the element count. Not what
      * {@link #getArrayInfo()} reports - see there.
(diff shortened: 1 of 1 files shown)


chrisdutz:
May I ask what the intention is behind PRs like this?
Yes: It might save a bit of CPU time, but it increases the memory footprint a tiny bit too ... 
Are you trying to solve a real problem or are you simply looking for possible contributions?

andvasp:
Hi @chrisdutz.

It’s not a critical issue; since I’ve been running a lot of tests and working on the EIP implementation, I found a simple improvement—especially since that method is called on every request, and I thought it wouldn't be too much work for you review it.

I am indeed trying to contribute to the project because it’s something that’s useful for my own work. The range of protocols supported in Java is fantastic!

I reported several test results regarding the refactoring of the SPI3 for S7, AB, and EIP in an effort to help out. My name won't appear in the codebase for those specific cases.

When I worked on the Connection Manager detection, it was a real issue for me because the read operations were very slow and could potentially impact the PLC itself. My goal wasn't just to become "another contributor."

I completely respect the project's design choices regarding CPU vs. memory footprint. If you feel this kind of micro-optimization isn't worth the memory trade-off, I'm more than happy to close this PR. Thanks for taking the time to review!

Best regards,
Anderson

chrisdutz:
Hi Anderson,

I hope I didn't offend you by my question ... it's just that I've been reviewing loads of PRs lately and even more that you don't even get to see and I am happy to review things where someone truly cares about the project. Also please see the latest update to the "Contributing" page which will become the official version after the release: 
https://plc4x.apache.org/plc4x/pre-release/developers/contributing.html

I've added a block about AI PRs and what we will be expecting of people.

I think you are right ... the tag will be parsed once and then potentially executed millions of times. I think it's ok to merge such a change.

Thanks for taking the time to explain.

Chris

chrisdutz (approved):

andvasp:
Thank you Chris.

Yes. I can imagine that a lot of work to do the reviews. The job the team is doing here is wonderful.

Again: This project is fantastic. The ability to use a lot of protocols with just one code base in Java is great.

I will introduce myself on the mailling list as soon as I get some time.

PR #2734, opened 2026-09-06
fix(plc4j/spi): RequestThrottle silently drops a limit reduction while requests are in flight

alxkm:
`RequestThrottle.adjustMaxConcurrentRequests()` loses the part of a reduction it cannot apply while requests are in flight, leaving the throttle permanently above its configured limit. This change defers that part as a permit debt that is settled as the running requests complete.

## Problem

`adjustMaxConcurrentRequests()` took permits away with

    int toDrain = Math.min(-difference, semaphore.availablePermits());

and then updated `maxConcurrentRequests` unconditionally. A permit that is in flight cannot be taken back, so whenever more requests were running than the new limit allows, the clamped-away part of the reduction was silently discarded while the field claimed it had been applied. As the running requests returned their permits, the semaphore climbed back to the old count and stayed there: the throttle permanently allowed more concurrent requests than it was configured for, which is the one thing the class exists to prevent. Repeated adjustments compound it - shrinking under load and raising the limit again leaves the pool over budget for good.

This is not limited to the case where no permit is free. With 3 permits and 2 in flight, reducing to 1 takes the single free permit and drops the second; `getInFlightRequests()` then reports 1 while 2 requests are running, and -1 once they finish.

## Fix

The part of a reduction that cannot be applied immediately is remembered as a permit debt and settled out of the permits the running requests return, so the reduction takes effect as soon as it can instead of being lost. Raising the limit cancels outstanding debt before handing out new permits. The adjustment log line reports how much of a reduction is still outstanding.

Reducing no longer blocks. The old `acquireUninterruptibly()` could wait, under the instance monitor, for permits that only the in-flight requests could hand back, because `availablePermits()` may drop between the check and the call. Free permits are now taken with non-blocking `tryAcquire()`, and whatever is left becomes debt.

`getInFlightRequests()` now counts the permits handed out directly instead of deriving the number from the ceiling and the free permits. The derived value could read negative not only after a lost reduction but also while the limit was being raised, because the permits were released before the ceiling was updated. While a debt is outstanding the counter may temporarily exceed `getMaxConcurrentRequests()`; that is the number of requests actually running, and it settles as they complete.

## Tests

The 15 existing tests are unchanged and keep passing. The 8 added tests all fail without this change. A package-private `getPermitDebt()` exposes the outstanding debt to the tests.

## Reachability

`RequestThrottle` is a `final` field on `ConnectionBase`, so it outlives any number of `connect()` calls, and `connect()` has no re-entry guard: an application-driven reconnect re-runs S7's `SetupCommunication` and adjusts a throttle that is already serving traffic. No driver triggers this on its own today, so this is a latent defect rather than an observed production failure.

## Alternative considered

A larger change would drop the semaphore from the `execute()` path in favour of an `AtomicInteger` in-flight counter checked against a `volatile` maximum, which would make a reduction a plain assignment. That touches the public `acquire()`/`release()` pair - unused elsewhere in the tree, but public API - so it is left out of this change.

### DIFF
--- a/plc4j/spi/drivers/src/main/java/org/apache/plc4x/java/spi/drivers/throttle/RequestThrottle.java
+++ b/plc4j/spi/drivers/src/main/java/org/apache/plc4x/java/spi/drivers/throttle/RequestThrottle.java
@@ -26,6 +26,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
 /**
@@ -41,6 +42,10 @@ public class RequestThrottle {
     /** Requests waiting for a permit, started in the order they were submitted. */
     private final Queue<Runnable> pending = new ConcurrentLinkedQueue<>();
     private final AtomicBoolean draining = new AtomicBoolean();
+    /** Part of a reduction that could not be applied at once, settled as in-flight permits return. */
+    private final AtomicInteger permitDebt = new AtomicInteger();
+    /** Permits handed out to requests that have not returned them yet. */
+    private final AtomicInteger inFlight = new AtomicInteger();
 
     public RequestThrottle(int maxConcurrentRequests) {
         if (maxConcurrentRequests < 1) {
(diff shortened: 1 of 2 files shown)


ottlukas:
Hi @alxkm ,

Thanks for opening this Pull Request.

Per our updated contribution guidelines, we ask that contributors who plan to introduce substantial changes or new capabilities—such as work touching our Serial Peripheral Interface (SPI)—first introduce themselves on the dev mailing list (dev@plc4x.apache.org).

Could you please post a brief introduction on the mailing list covering:

A quick introduction of yourself and your context.

Your goals and intent for this PR, specifically what you are looking to achieve with these changes to the SPI. (Your use case and machines etc.)

Open-source maintainability relies heavily on context and architectural alignment before reviewing code diffs. Introducing yourself on the list gives the broader community a chance to understand the use case, offer early architectural feedback, and coordinate efforts effectively.

Once you've posted to the list, please link the thread or drop a note here so we can align on the proposed changes before proceeding with the code review.

Thanks!

This helps us to assure maintainability and is a way against AI Slop protection with Pull requests just opening but no intent on further contributions.

alxkm:
Hi @ottlukas

Thanks - makes sense. Posted an introduction and the context behind this PR
to the dev list: https://lists.apache.org/thread/4y0c7omcydg1b7szcnw5wm3tswq0v3tl

Short version: no hardware and no production use case behind it, I came to
this from the JVM side. Happy to align on the approach there before the code
review, and to rework or split the change if that fits better.

PR #2761, opened 2026-09-20
chore(plc4net): remove the abandoned net452 knxproj driver

TomNewChao:
`plc4net/drivers/knxnetip` contains `plc4net-driver-knxproj` — a .NET Framework
4.5.2 parser for ETS `.knxproj` project files. Despite the directory name it is
not a KNXnet/IP protocol driver. It targets `net452`, so it cannot build on the
Linux or macOS CI runners, and it has never been part of a working plc4net build.

This has to go before `api/` and `spi/` retarget to net8.0, not just for
readability — it would stop building. `plc4net-driver-knxproj.csproj` references
`api/plc4net-api.csproj` and `spi/plc4net-spi.csproj` directly, and a `net452`
project cannot reference a `net8.0`-only library. The next PR in this series
(decomposing #2656) does that retarget; if this driver were still around when it
lands, `develop` would sit on a broken build between the two merges. Splitting it
out here also keeps that PR reviewable on its own terms: carried together, this
deletion dominates the diff at roughly 45k lines and buries the ~9k lines that
are the actual net8.0 port. Nothing outside `plc4net/` referenced it, so it is
reviewable at a glance on its own — a pure deletion.

Along with the driver, and nothing else:

- the two project entries, their build configurations and the now-empty
  `drivers` solution folder in `plc4net.sln`
- the `maven-dependency-plugin` execution that unpacked the `knxnetip` protocol
  test-jar into `drivers/knxnetip-test/resources`, and the test-jar dependency it
  needed
- the `update-generated-code` profile, whose only execution generated this driver
  through the Java `plc4x-maven-plugin` with `languageName=C#`
- the dead `/drivers/knxnetip/bin/` entry in `.gitignore`

**187 files changed, 45,326 deletions, 0 insertions.**

### What this does not do

It does not replace the removed functionality. The revival provides a KNXnet/IP
*protocol* driver, which is a different thing — `.knxproj` file parsing is simply
gone from plc4net after this. If it is wanted in .NET, it should come back as its
own contribution against the current runtime rather than as net452 code nobody
can build.

It also does not fix the plc4net build: `api/` and `spi/` are still `net452`
after this change. That is the next PR's job, described above. `plc4net` builds
only under the `-Pwith-dotnet` profile and is not part of the default reactor, so
nothing in the default build is affected either way.

### Context

This is the first of a series of PRs decomposing #2656, which is now a draft
spike rather than a merge candidate. The proposed order, and the reasoning behind
it, is in that PR's description. Feedback on the order is welcome there.


TomNewChao:
@sruehl 
The end-to-end work in #2656 has now been validated against real hardware:
  - S7 was verified against a Siemens S7-1214C.
  - Modbus RTU wire exchange, framing, CRC, and the driver read path were verified over a real serial link using a Mitsubishi QJ71C24N as a fixed-response slave.

The detailed procedures, results, and limitations are documented here:
    https://github.com/openIndu/plc4x/blob/feature/plc4net-revival/plc4net/docs/hardware-verification.md

Now that the implementation path has been proven, I am splitting #2656 into smaller, dependency-ordered PRs, following your review-size feedback. The spike has grown to hundreds of files and contains large generated trees, so reviewing or merging it as one change would not be practical.

#2761 is the first PR in that decomposition. It removes only the abandoned net452 .knxproj driver and its dedicated build wiring: 187 files and 45,326 lines deleted, with no insertions.

This deletion comes first because:
  1. The old net452 project cannot reference the net8.0 API and SPI introduced by the next PR.
  2. Keeping the deletion in the next PR would bury the actual net8.0 implementation under roughly 45,000 deleted lines.
  3. Landing it separately keeps each following PR focused and reviewable.

This first PR is ready for review. When you have time, could you please review it and let me know whether this decomposition direction matches what you had in mind?

Thanks

TomNewChao:
@chrisdutz, while updating this branch to the latest develop, I noticed that the recent KNX vendor refresh also updated KnxManufacturer.cs in the legacy plc4net .knxproj parser removed by this PR.

My understanding is that this net452 project is not part of the current working plc4net build and cannot remain unchanged once the API and SPI move to net8.0. I separated its removal because its generated sources would otherwise dominate the actual migration diff.

Before proceeding, could you clarify whether this parser is still considered useful functionality that should be preserved and ported? If so, I can revise the plan instead of treating it as abandoned code.

splatch:
I admitted workflow run, but it seems to be stuck in queue. Lets see how long it will be there before timeout. I know that this PR is harmless, so we can evaluate just execution permissions for github workflow(s). Which is a mystery to me.

TomNewChao:
> I admitted workflow run, but it seems to be stuck in queue. Lets see how long it will be there before timeout. I know that this PR is harmless, so we can evaluate just execution permissions for github workflow(s). Which is a mystery to me.

Thanks @splatch for approving the workflow.

The run eventually completed, and the Dependency Review check is now passing. GitHub currently shows the PR as mergeable, with the remaining visible gate being one approving review from a reviewer with write access.

Since you mentioned that the change is harmless, would you be willing to submit a formal review when convenient? If the plc4net/KNX part should be reviewed by someone else, please let me know who would be the most appropriate reviewer.

Thanks again,
Tom

splatch (approved):

